Issue #97, opened 2023-11-28
Fix Wave Terminal Icon for Linux


sawka:
No Icon is showing up for Linux users in the dock.

It looks like we need a different icon format for Linux / Windows builds:
https://stackoverflow.com/questions/31529772/how-to-set-app-icon-for-electron-atom-shell-app

Right now we're just setting the icon in "icns" format (which only seems to work for MacOS).

nathanv:
Better docs here: https://www.electronforge.io/guides/create-and-add-icons#linux

sawka:
This should be fixed with v0.5.1

nathanv:
I am still having issues. I upgrade to 0.5.1 but the dot files are from 0.5.0 . Is there anything I need to do to fix?

PR #120, opened 2023-12-04
Add Waveterm Icon to Linux

oneirocosm:
Issue #97 brought up that the icon does not appear on linux.  This change provides the wave-logo-dark.png to the Electron Browser Window to fix that.


sawka:
Ah, we should only conditionally add this if the platform is linux.  We don't want this icon to override the "icns format" MacOS icon.  can check for `unamePlatform == "linux"`

PR #124, opened 2023-12-05
migrate screen settings modal to new modals system

adred:


src/app/common/common.tsx
```diff
@@ -199,7 +200,7 @@ class Tooltip extends React.Component<TooltipProps, TooltipState> {
         const style = this.calculatePosition();
 
         return ReactDOM.createPortal(
-            <div className="wave-tooltip" style={style}>
+            <div className={`wave-tooltip ${this.props.className}`} style={style}>
```
sawka, line 203: change to: cn("wave-tooltip", this.props.className)
(better because it handles the undefined/null case automatically without an `|| ""`)

src/app/common/common.tsx
```diff
@@ -1070,6 +1071,7 @@ class Dropdown extends React.Component<DropdownProps, DropdownState> {
             <div
                 className={cn(`wave-dropdown ${className || ""}`, {
```
sawka, line 1072: change this to cn("wave-dropdown", className), ...)
(no need for the `|| ""`) in that case

PR #138, opened 2023-12-13
Fix potential overflow in promptenc arithmetic

esimkowitz:
This fixes a potential integer overflow in how we calculate the buffer size for promptenc.EncryptData.

This introduces two new utilfns for detecting overflow when adding ints.

### DIFF
--- a/wavesrv/pkg/utilfn/utilfn.go
+++ b/wavesrv/pkg/utilfn/utilfn.go
@@ -6,6 +6,8 @@ package utilfn
 import (
 	"crypto/sha1"
 	"encoding/base64"
+	"errors"
+	"math"
 	"regexp"
 	"strings"
 	"unicode/utf8"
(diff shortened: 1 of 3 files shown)


wavesrv/pkg/utilfn/utilfn.go
```diff
@@ -209,3 +211,32 @@ func ChunkSlice[T any](s []T, chunkSize int) [][]T {
...
+var ErrOverflow = errors.New("integer overflow")
+
+// Add two int values, returning an error if the result overflows.
+func AddInt(left, right int) (int, error) {
+	if right > 0 {
+		if left > math.MaxInt-right {
+			return 0, ErrOverflow
+		}
+	} else {
+		if left < math.MaxInt-right {
```
sawka, line 224: should this be math.MinInt?

esimkowitz: good catch

PR #139, opened 2023-12-13
Clean up potentially unsafe external links

esimkowitz:
Addresses some flagged "vulnerabilities" in our React code. It's not as much of an issue since all links will open in the browser and the `window.Opener` object should never be accessible, but it's still good practice. I am also cleaning up some other linting warnings that I came across.


sawka:
looks good so far!

esimkowitz:
Making a draft again as there's a bug in the connections page now, I think some of my linting changes might not be working right. Will look again in the morning

esimkowitz:
Fixed the broken imports in common.tsx, figured regressions in infomsg.tsx conditionals

Issue #99, opened 2023-11-28
Ignore spurious SSH output that precedes init packet 


sawka:
from discord.  a user reported a broken remote connection because their SSH implementation outputs a special string at the beginning of the SSH session.  the init packet follows the extra output.

maybe it is possible to ignore spurious output on the connection until we see the "\n##" start of the init packet (and not throw an error)?

adding this bug as a placeholder to investigate whether adding this fix will be easy or not.

PR #140, opened 2023-12-13
Packetparser Ignore Spurious Invalid Input at Beginning of Stream

sawka:
Possible fix for #99.  Add a flag to have the parser ignore spurious input that happens before seeing a real packet.  Normally that spurious input is converted in a RawPacket.  Those RawPackets cause connection errors (since the code on the other side is waiting for an InitPacket).  This flag will let the parser ignore input until it sees a valid packet.

We also need to force all of the paths that use this new flow to output a valid packet.  That's why I added the "ping" packets in the mshell commands.

### DIFF
--- a/waveshell/main-waveshell.go
+++ b/waveshell/main-waveshell.go
@@ -156,7 +156,7 @@ func readFullRunPacket(packetParser *packet.PacketParser) (*packet.RunPacketType
 }
 
 func handleSingle(fromServer bool) {
-	packetParser := packet.MakePacketParser(os.Stdin, false)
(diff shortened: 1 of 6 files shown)


sebastiangebhardt:
Short, question. I have an ssh server configured with a banner message at connect. I get an error like this:
`prompt> error connecting to remote: invalid packet received from mshell client: raw[=== BANNER TEXT===]`
`prompt> error reading from controlling-pty: read /dev/ptmx: input/output error`
Maybe this is fixed by this pull request?

Issue #133, opened 2023-12-11
Specify OpenAI API Base url


shuther:
**Is your feature request related to a problem? Please describe.**
I would like to use self hosted model or to leverage Azure endpoint.

**Describe the solution you'd like**
Be able to specify the openAI base url.

sawka:
Thanks for your submission!  This makes a lot of sense and should be straightforward to implement.

sawka:
For our own reference, linking to the docs for how to provide the config for setting the baseurl for the package we're using:
https://pkg.go.dev/github.com/sashabaranov/go-openai#ClientConfig

PR #141, opened 2023-12-13
Allow user to specify openAI baseURL

MrStashley:
PR for https://github.com/wavetermdev/waveterm/issues/133

Exposes open ai BaseURL in client:set command

User can set BaseURL with `/client:set openaibaseurl=<BASEURL>`

--TESTING--
I couldn't find any other hosted open ai models to test this with, but I was able to confirm with logs that the baseURL is correctly plumbed through the client data opts and it appears correctly when used by the chat command. I was also able to set the base url to the default open ai base url and verify that the chat command still worked, as well as verify that the chat command stopped working when given a bogus base url 

Screen Shot 2023-12-12 at 5 15 13 PM

--CONSIDERATIONS--
Something to consider about this pr: The go openai library that we use has a special Azure type. if azure baseUrl is specified, and APITypeAzure is not specified, client.fullURL is not going to work correctly. This shouldn't be a big deal, we don't currently use client.fullURL(), but it's something to note if we need it in the future. 

Also authenticating with Azure API Keys is currently not supported, must authenticate with Microsoft Entra authentication, which works exactly the same as Open AI token authentication.

### DIFF
--- a/wavesrv/pkg/cmdrunner/cmdrunner.go
+++ b/wavesrv/pkg/cmdrunner/cmdrunner.go
@@ -3639,8 +3639,21 @@ func ClientSetCommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (ss
 			return nil, fmt.Errorf("error updating client openai maxchoices: %v", err)
 		}
 	}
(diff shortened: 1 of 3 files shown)


sawka:
this looks great!

PR #148, opened 2023-12-15
Add support for proxying open AI chat completion through cloud

MrStashley:
Screen Shot 2023-12-14 at 8 02 52 PM
Added support for proxying open ai chat completion through cloud, this allows users to not have to specify their own openai token

If token is not set, chat completion will now be proxied through the cloud. You can unset your token through the UI. 
the cloud uses the default model, currently "gpt-3.5-turbo".
It allows you to set max tokens and max choices parameter

If you don't have telemetry set to on, you cannot use this feature 

-- CONSIDERATIONS --
For now, this is set to only hit our dev lambda server. When we deploy the completion code to our prod lambda server, we should add a check for is dev or prod

-- TESTING --
I tested with telemetry on and off, and confirmed that the correct error shows. I have tested the channel timeout and confirm that if the socket hangs, the channel will timeout and send an error, and then the terminal can continue as normal.

I have tested all of the standard code paths, confirmed that when token is set we will still do completion locally, and when token is not set we will do cloud completion. I confirmed that token can now be set to empty string, and it looks good in the ui (it goes back to "not set"). I have also checked socket failed to connect error.

There are some more unlikely error paths that I haven't checked because I didn't reproduce that error, like json decode, etc etc. I also haven't tested socket connect timeout error, but I expect these paths to work as described because the behavior is fairly standard

What do we think about the way I do testing? I can write unit tests in the future

### DIFF
--- a/wavesrv/pkg/cmdrunner/cmdrunner.go
+++ b/wavesrv/pkg/cmdrunner/cmdrunner.go
@@ -22,6 +22,7 @@ import (
 	"unicode"
 
 	"github.com/google/uuid"
+	"github.com/gorilla/websocket"
 	"github.com/wavetermdev/waveterm/waveshell/pkg/base"
 	"github.com/wavetermdev/waveterm/waveshell/pkg/packet"
 	"github.com/wavetermdev/waveterm/waveshell/pkg/shexec"
@@ -1550,7 +1551,9 @@ func doOpenAIStreamCompletion(cmd *sstore.CmdType, opts *sstore.OpenAIOptsType, 
 			writeErrorToPty(cmd, fmt.Sprintf("In order to protect against abuse, you must have telemetry turned on in order to use Wave's free AI features.  If you do not want to turn telemetry on, you can still use Wave's AI features by adding your own OpenAI key in Settings.  Note that when you use your own key, requests are not proxied through Wave's servers and will be sent directly to the OpenAI API."), outputPos)
 			return
 		}
-		ch, err = openai.RunCloudCompletionStream(ctx, opts, prompt)
+		var conn *websocket.Conn
+		ch, conn, err = openai.RunCloudCompletionStream(ctx, opts, prompt)
+		defer conn.Close()
 	} else {
 		ch, err = openai.RunCompletionStream(ctx, opts, prompt)
 	}
--- a/wavesrv/pkg/cmdrunner/cmdrunner.go
+++ b/wavesrv/pkg/cmdrunner/cmdrunner.go
@@ -68,6 +68,7 @@ const TermFontSizeMax = 24
 const TsFormatStr = "2006-01-02 15:04:05"
 
 const OpenAIPacketTimeout = 10 * time.Second
+const OpenAIStreamTimeout = 60 * time.Second
 
 const OpenAICloudCompletionTelemetryOffErrorMsg = "In order to protect against abuse, you must have telemetry turned on in order to use Wave's free AI features.  If you do not want to turn telemetry on, you can still use Wave's AI features by adding your own OpenAI key in Settings.  Note that when you use your own key, requests are not proxied through Wave's servers and will be sent directly to the OpenAI API."
 
@@ -1511,7 +1512,7 @@ func doOpenAIStreamCompletion(cmd *sstore.CmdType, opts *sstore.OpenAIOptsType, 
 	var outputPos int64
 	var hadError bool
 	startTime := time.Now()
-	ctx, cancelFn := context.WithTimeout(context.Background(), 30*time.Second)
+	ctx, cancelFn := context.WithTimeout(context.Background(), OpenAIStreamTimeout)
 	defer cancelFn()
 	defer func() {
 		r := recover()
@@ -1561,8 +1562,9 @@ func doOpenAIStreamCompletion(cmd *sstore.CmdType, opts *sstore.OpenAIOptsType, 
 		select {
 		case <-time.After(OpenAIPacketTimeout):
 			// timeout reading from channel
-			timeoutPk := openai.CreateTextPacket(fmt.Sprintf("... (timed out waiting for server response)"))
-			err = writePacketToPty(ctx, cmd, timeoutPk, &outputPos)
+			hadError = true
+			pk := openai.CreateErrorPacket(fmt.Sprintf("timeout waiting for server response"))
+			err = writePacketToPty(ctx, cmd, pk, &outputPos)
 			if err != nil {
 				log.Printf("error writing response to ptybuffer: %v", err)
 				return
@@ -1572,8 +1574,12 @@ func doOpenAIStreamCompletion(cmd *sstore.CmdType, opts *sstore.OpenAIOptsType, 
 		case pk, ok := <-ch:
 			if ok {
 				// got a packet
+				if pk.Error != "" {
+					hadError = true
+				}
 				err = writePacketToPty(ctx, cmd, pk, &outputPos)
 				if err != nil {
+					hadError = true
 					log.Printf("error writing response to ptybuffer: %v", err)
 					return
 				}
--- a/wavesrv/pkg/cmdrunner/cmdrunner.go
+++ b/wavesrv/pkg/cmdrunner/cmdrunner.go
@@ -1561,7 +1561,7 @@ func doOpenAIStreamCompletion(cmd *sstore.CmdType, opts *sstore.OpenAIOptsType, 
 		select {
 		case <-time.After(OpenAIPacketTimeout):
 			// timeout reading from channel
-			timeoutPk := openai.CreateErrorPacket(fmt.Sprintf("Server timed out waiting for packets"))
+			timeoutPk := openai.CreateTextPacket(fmt.Sprintf("... (timed out waiting for server response)"))
 			err = writePacketToPty(ctx, cmd, timeoutPk, &outputPos)
 			if err != nil {
 				log.Printf("error writing response to ptybuffer: %v", err)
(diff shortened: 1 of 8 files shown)


wavesrv/pkg/remote/openai/openai.go
```diff
@@ -68,6 +74,96 @@ func RunCompletion(ctx context.Context, opts *sstore.OpenAIOptsType, prompt []ss
...
+		log.Printf("Websocket error: %v", err)
+		return nil, fmt.Errorf("Websocket error: %v", err)
+	}
+	cloudCompletionRequestConfig := sstore.OpenAICloudCompletionRequest{
+		Model:      opts.Model,
+		Prompt:     prompt,
+		MaxTokens:  opts.MaxTokens,
+		MaxChoices: opts.MaxChoices,
+	}
+	configMessageBuf := new(bytes.Buffer)
```
MrStashley, line 135: should I change this to marshal instead of decode?

wavesrv/pkg/remote/openai/openai.go
```diff
@@ -68,6 +74,96 @@ func RunCompletion(ctx context.Context, opts *sstore.OpenAIOptsType, prompt []ss
...
 
+func RunCloudCompletion(ctx context.Context, opts *sstore.OpenAIOptsType, prompt []sstore.OpenAIPromptMessageType) ([]*packet.OpenAIPacketType, error) {
+	if opts == nil {
+		return nil, fmt.Errorf("no openai opts found")
+	}
+	if opts.Model == "" {
+		return nil, fmt.Errorf("no openai model specified")
+	}
+	cloudCompletionRequestConfig := sstore.OpenAICloudCompletionRequest{
+		Model:      opts.Model,
```
MrStashley, line 85: currently sends model to the server - we shouldn't do this if we don't plan to use it

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1445,7 +1445,16 @@ func doOpenAICompletion(cmd *sstore.CmdType, opts *sstore.OpenAIOptsType, prompt
 		}
 		sstore.MainBus.SendScreenUpdate(cmd.ScreenId, update)
 	}()
-	respPks, err := openai.RunCompletion(ctx, opts, prompt)
+	var respPks []*packet.OpenAIPacketType
+	var err error
+	log.Printf("TODO: fix this condition - short circuited to access cloud code path\n")
```
MrStashley, line 1450: Here's where I short circuit the code path so it always proxies through cloud

wavesrv/pkg/remote/openai/openai.go
sawka: agree, we should hard code the model on the server side

wavesrv/pkg/remote/openai/openai.go
sawka: ya, it is easier that way.  Marshal will give you a []byte directly, then you don't even need the bytes.Buffer.

wavesrv/pkg/remote/openai/openai.go
```diff
@@ -68,6 +74,96 @@ func RunCompletion(ctx context.Context, opts *sstore.OpenAIOptsType, prompt []ss
...
+		MaxChoices: opts.MaxChoices,
+	}
+	configMessageBuf := new(bytes.Buffer)
+	json.NewEncoder(configMessageBuf).Encode(cloudCompletionRequestConfig)
+	err = conn.WriteMessage(websocket.TextMessage, configMessageBuf.Bytes())
+	if err != nil {
+		return nil, fmt.Errorf("Websocker write config error: %v", err)
+	}
+	rtn := make(chan *packet.OpenAIPacketType, DefaultStreamChanSize)
+	go func() {
```
sawka, line 142: we should think about how to timeout this conn/request if something goes wrong or something hangs.  either a hard timeout after a certain amount of time, or a rolling timeout, like if we don't get a new packet for X seconds, then we kill the connection and exit the loop.  the "killer" can run in a different go routine (as ReadMessage can hang).

wavesrv/pkg/remote/openai/openai.go
```diff
@@ -68,6 +74,96 @@ func RunCompletion(ctx context.Context, opts *sstore.OpenAIOptsType, prompt []ss
...
+
+func RunCloudCompletionStream(ctx context.Context, opts *sstore.OpenAIOptsType, prompt []sstore.OpenAIPromptMessageType) (chan *packet.OpenAIPacketType, error) {
+	const AWSLambdaCentralWSAddr = "wss://5lfzlg5crl.execute-api.us-west-2.amazonaws.com/dev/"
+	if opts == nil {
+		return nil, fmt.Errorf("no openai opts found")
+	}
+	if opts.Model == "" {
+		return nil, fmt.Errorf("no openai model specified")
+	}
+	conn, _, err := websocket.DefaultDialer.Dial(AWSLambdaCentralWSAddr, nil)
```
sawka, line 124: Use DialContext and we should set some sensible timeout value for connecting

sawka:
I'm thinking we should completely remove the non-streaming paths in the code.  Not sure they are really useful.  Can't think of a time when we wouldn't want to stream the results back?

wavesrv/pkg/remote/openai/openai.go
MrStashley: addressed in doOpenAIStreamCompletion - cmdrunner.go:1526

I added a timeout to the channel, if we don't get a new message sent from the socket read goroutine in enough time, the function sends a timeout error

what do you think about this?
Some issues with this are if the rehdmessage call hangs, this doesn't kill the goroutine bc it's somewhat difficult to do so, so that process will persist taking up resources. If there's a way to set a timeout to web socket read message, I should do that as well 

Also, it sends a timeout error packet so the client loses all the information it had up until the timeout. I think it might be better to add a packet to the output that looks like [Server timeout] or something, so the client can still see what they got up to that point. That's difficult to do tho, because the openai packets have an index and other stuff. We would have to save the last packet in order to create our own

What do you think? is this okay for this release or should I revisit this?

wavesrv/pkg/cmdrunner/cmdrunner.go
MrStashley: addressed

wavesrv/pkg/remote/openai/openai.go
MrStashley: Addressed, set a 20 second timeout

MrStashley:
High latency environments might want non streaming completion?
But even then streaming might be better?

web security considerations?
Not sure

wavesrv/pkg/remote/openai/openai.go
MrStashley: addressed

wavesrv/pkg/remote/openai/openai.go
```diff
@@ -72,6 +79,60 @@ func RunCompletion(ctx context.Context, opts *sstore.OpenAIOptsType, prompt []ss
...
+	}
+	websocketContext, _ := context.WithTimeout(context.Background(), CloudWebsocketConnectTimeout)
+	conn, _, err := websocket.DefaultDialer.DialContext(websocketContext, AWSLambdaCentralWSAddr, nil)
+	if err != nil {
+		log.Printf("Websocket error: %v", err)
+		return nil, fmt.Errorf("Websocket error: %v", err)
+	}
+	cloudCompletionRequestConfig := sstore.OpenAICloudCompletionRequest{
+		Prompt:     prompt,
+		MaxTokens:  opts.MaxTokens,
```
MrStashley, line 98: I still let the user specify max tokens and max choices for the cloud, should I use the defaults instead or is this ok?

sawka: I think we shouldn't let the user specify these.  We should just set them in the cloud.

wavesrv/pkg/remote/openai/openai.go
sawka: The way to "unstick" a file/connection read in go is to close the connection.  So once you do websocketConnection.Close() (or whatever the gorilla equivalent is) that should cause the readmessage call to fail.  (be careful about setting two error messages in this case though)

wavesrv/pkg/remote/openai/openai.go
MrStashley: addressed

wavesrv/pkg/remote/openai/openai.go
MrStashley: addressed. My solution for this is to pass the web socket connection as a parameter to the channel read function and close the socket when it is done, so the socket always closes. This shouldn't have any issues with double errors because the reader stops reading when it hits its timeout

The gorilla web socket documentation didn't say whether it was ok to call close on a socket twice (in the best case the channel write goroutine and the channel read goroutine both call close one after the other), but I tested it and didn't have any issues

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1533,16 +1539,54 @@ func doOpenAIStreamCompletion(cmd *sstore.CmdType, opts *sstore.OpenAIOptsType,
...
+	var ch chan *packet.OpenAIPacketType
+	var err error
+	if opts.APIToken == "" {
+		clientData, err := sstore.EnsureClientData(ctx)
+		if err != nil {
+			writeErrorToPty(cmd, fmt.Sprintf("error getting client data: %v", err), outputPos)
+			return
+		}
+		if clientData.ClientOpts.NoTelemetry {
+			writeErrorToPty(cmd, fmt.Sprintf("In order to protect against abuse, you must have telemetry turned on in order to use Wave's free AI features.  If you do not want to turn telemetry on, you can still use Wave's AI features by adding your own OpenAI key in Settings.  Note that when you use your own key, requests are not proxied through Wave's servers and will be sent directly to the OpenAI API."), outputPos)
```
MrStashley, line 1551: Long message, should this be a const?

In the future should there be a strings / error msgs package?

MrStashley: made it a const and put it at the top of cmdrunner.go file

PR #182, opened 2023-12-20
ScreenTabs refactor

adred:
This PR includes

- Fix UI jump reported by @esimkowitz
- Fix reorder not working from CLI
- Move out renderTab() as a separate component
- General improvements

### DIFF
--- a/src/app/workspace/screen/tab.tsx
+++ b/src/app/workspace/screen/tab.tsx
@@ -25,7 +25,10 @@ import "./tabs.less";
 dayjs.extend(localizedFormat);
 
 @mobxReact.observer
-class ScreenTab extends React.Component<{ screen: Screen; activeScreenId: string; index: number; onSwitchScreen }, {}> {
+class ScreenTab extends React.Component<
+    { screen: Screen; activeScreenId: string; index: number; onSwitchScreen: (screenId: string) => void },
+    {}
+> {
     tabRefs: { [screenId: string]: React.RefObject<any> } = {};
     dragEndTimeout = null;
     scrollIntoViewTimeout = null;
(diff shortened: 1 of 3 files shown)


adred:
This no longer converts `Screens[]` to plain JS. Triggering of re-render when `any` occurs in `Screens[]` is captured via `mobx.reaction`.

src/app/workspace/screen/tab.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { MagicLayout } from "../../magiclayout";
+
+import "../workspace.less";
+import "./tabs.less";
+
+dayjs.extend(localizedFormat);
+
+@mobxReact.observer
+class ScreenTab extends React.Component<{ screen: Screen; activeScreenId: string; index: number; onSwitchScreen }, {}> {
+    tabRefs: { [screenId: string]: React.RefObject<any> } = {};
```
sawka, line 29: why do we need (potentially) multiple refs here?  The same ScreenTab can be used by different screenIds?

sawka: because the ScreenTab `key` is set to screen.screenId we should get a new component anytime the screenId changes (unless drag&drop is doing something weird)

src/app/workspace/screen/tabs.tsx
```diff
@@ -267,7 +211,13 @@ class ScreenTabs extends React.Component<{ session: Session }, { showingScreens:
                 >
                     <For each="screen" index="index" of={showingScreens}>
                         <React.Fragment key={screen.screenId}>
```
sawka, line 213: Because ScreenTab is now a component, this `<React.Fragment>` can be removed.

src/app/workspace/screen/tab.tsx
adred: refs are used to determine the new tab index to be passed to the backend, based on their position in the DOM. Please see the `handleDragEnd` method for reference.

PR #183, opened 2023-12-20
Clean up cirfile and cirfile_tests, ensure we are using contexts with cancellation signals

esimkowitz:
I've cleaned up some other linting warnings in `cirfile.go` and `cirfile_tests.go`, notably the use of nil contexts. I've worked to ensure that where file operations could indefinitely block a routine (such as not having a cancellation or timeout), we return EWOULDBLOCK.

### DIFF
--- a/waveshell/pkg/cirfile/cirfile.go
+++ b/waveshell/pkg/cirfile/cirfile.go
@@ -102,6 +102,8 @@ var cfRegex = regexp.MustCompile(`([^.\v]+[\\|\/])+([^.\v]+.cf)`)
 var tempDir = os.TempDir()
 var waveHomeDir = scbase.GetWaveHomeDir()
 
+const unableToValidateCirFilePath = "unable to validate cirfile path[%s]: %w"
+
 // returns error if the filename is not a valid cirfile path
 func ValidateCirFilePath(fileName string) error {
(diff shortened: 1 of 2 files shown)


esimkowitz:
I've removed my validate path change as @sawka correctly indicated that all inputs to `cirfile` are already validated and as such can be considered "trusted". I'll mark the security alert as a false alarm

waveshell/pkg/cirfile/cirfile.go
```diff
@@ -148,14 +150,16 @@ func CreateCirFile(fileName string, maxSize int64) (*File, error) {
 		return nil, err
 	}
 	rtn := &File{OSFile: fd, Version: CurrentVersion, MaxSize: maxSize, StartPos: FilePosEmpty}
-	err = rtn.flock(nil, syscall.LOCK_EX)
+	timeoutContext, cancelFn := context.WithTimeout(context.Background(), 20*time.Millisecond)
```
sawka, line 153: i'm going to change this back to passing `nil`.  I think that's the right call here.  This is a new file.  The only way that the flock could block is if someone else is creating this exact same file at the same time.  In that situation it would be better to error immediately (with the EWOULDBLOCK) than to wait the 20ms and get the lock and proceed with the code.

PR #185, opened 2023-12-20
Escape error output in main-server to quiet security alert

esimkowitz:
We got an alert that we were not properly escaping an error output that we return when reading PtyOut. This could potentially expose XSS depending on what was in the request. This is extremely unlikely, but is an easy fix. I took this time to clean up some other linting errors, mainly around reusing const strings and unnecessary use of sprintf.


sawka:
Look good.  I'm going to cut the v0.5.3 release (to fix the workspace delete bug) first before pulling these fixes in.

PR #187, opened 2023-12-21
Immediately hide modal when deleting workspace

adred:


sawka:
We don't always want to close the modal.  We only want to close the modal if the delete happens without an error.

You can use prtn to see if there was an error like this:

```
prtn.then((rval) => {
    if (rval?.success) {
        // close the modal here
    }
});
```

PR #194, opened 2023-12-24
Tabs freeze fix

adred:
Previously, when swiping rapidly through over a hundred tabs, ScreenTabs would freeze and become unusable. This patch fixes that issue.


adred:
This is not a good fix.

PR #198, opened 2023-12-26
Fix Commands that Require Quoted Paths

oneirocosm:
Several commands did not wrap the path in quotes which caused problems when attempting to store the waveterm installation in a place that had a space in the path. This corrects this in the particular case where the username does not have spaces but the path to the executable does.

Note: the case of a user name having spaces has not been tested but likely does not work.


sawka:
The quoting is good, but it looks like you changed the the `bash -c` from calling `getWaveSrvCmd` to calling `getWaveSrvPath`.  So the server will no longer log to wavesrv.log.

PR #200, opened 2023-12-28
Cmd-P feature

adred:


src/app/common/modals/modals.less
```diff
@@ -453,6 +453,186 @@
...
+                        }
+                    }
+
+                    .selected-option {
+                        border: 1px solid rgba(241, 246, 243, 0.15);
+                        border-radius: 4px;
+                        background: rgba(255, 255, 255, 0.06);
+                    }
+
+                    .color-green,
```
sawka, line 530: i think all of these `color` classes should be combined into app.less.  we already have rules for .icon.color-X in there that will set the path fill color.  just add the `i { color: ...}` rules there as well.  (i think we also need color-default there as well).  then maybe add the `icon` class to your renderIcon output and we'll be set and can remove all of these duplicate classes.

adred: resolved

PR #207, opened 2023-12-30
Chat GPT Command Info Chat

MrStashley:
This feature creates a chat window that the user can open with Ctrl + Space that will allow them to query chatgpt for help on their current command. ChatGPT will know the command that they are running and will be able to help directly

I have a chat window that opens when the user presses ctrl + space, a functional chatgpt chat, and various qol and ux features, such as ability to easily capture code blocks, a clear hotkey, etc

Bugs
- seems like sometimes on the first ui action things can get messed up. I've noticed that sometimes the first time I clear the chat window the lines window doesn't get updated correctly, leaving a black empty space, which gets fixed next time the window updates. I also noticed that sometimes the first time I switch tabs the chat window doesn't change to reflect the data of the new screen. These errors are spurious and rare and I think they're below the level of my code, unless there are special events for first time actions. Should still investigate further, but I think these are rare enough to be non blocking, unless others disagree
- Mike mentioned he didn't get history correctly. I think this has been fixed but should verify before merging

### DIFF
--- a/src/app/common/common.tsx
+++ b/src/app/common/common.tsx
@@ -824,10 +824,13 @@ function HeaderRenderer(props: any, hnum: number): any {
     return <div className={cn("title", "is-" + hnum)}>{props.children}</div>;
 }
 
+function CodeRenderer(props:any): any {
+    return <code className={cn({ inline: props.inline })}>{props.children}</code>;
+}
+
 @mobxReact.observer
 class Markdown extends React.Component<{ text: string; style?: any; extraClassName?: string; codeSelect?: boolean}, {}> {
     codeBlockList: Array<React.RefObject<HTMLElement>>    
-    codeSelectMarkdownID: number;
 
     constructor(props: any) {
         super(props);  
@@ -835,24 +838,23 @@ class Markdown extends React.Component<{ text: string; style?: any; extraClassNa
         if(this.props.codeSelect) {
             let model = GlobalModel;
             let inputModel = model.inputModel; 
-            this.codeSelectMarkdownID = inputModel.registerCodeSelectMarkdown();   
         }
     }  
 
-   CodeRenderer(props: any, codeSelect: boolean): any { 
-        let codeText = props.children;
+    CodeBlockRenderer(props: any, codeSelect: boolean, ): any { 
+        let codeText = props.node.children[0].children[0].value;
         if(codeText) {
             codeText = codeText.replace(/\n$/, "") // remove trailing newline        
         }
-        let blockRef: React.RefObject<HTMLElement> = React.createRef();
+        let blockRef: React.RefObject<HTMLPreElement> = React.createRef();
         let clickHandler: (e: React.MouseEvent<HTMLElement>, blockIndex: number) => void;
         let curBlockIndex: number;
         if(codeSelect) {
             let model = GlobalModel;
             let inputModel = model.inputModel;
-            curBlockIndex = inputModel.addCodeBlockToCodeSelect(blockRef, this.codeSelectMarkdownID);
+            curBlockIndex = inputModel.addCodeBlockToCodeSelect(blockRef);
             clickHandler = (e: React.MouseEvent<HTMLElement>, blockIndex: number) => {
-                inputModel.setCodeSelectSelectedCodeBlock(blockIndex, this.codeSelectMarkdownID);
+                inputModel.setCodeSelectSelectedCodeBlock(blockIndex);
             };
         } else {
             clickHandler = (e: React.MouseEvent<HTMLElement>, blockIndex: number) => {
@@ -861,22 +863,9 @@ class Markdown extends React.Component<{ text: string; style?: any; extraClassNa
             curBlockIndex = this.addCodeBlockToList(blockRef);
         }
         return (
-            <code className={cn({ inline: props.inline})} ref={blockRef} onClick={(event) => clickHandler(event, curBlockIndex)}>{props.children}</code>
+            <pre ref={blockRef} onClick={(event) => clickHandler(event, curBlockIndex)}>{props.children}</pre>
         )
     
-    }
-    
-    setSelectedCodeBlock(selectIndex: number) {
-        for(let index = 0; index < this.codeBlockList.length; index++) {
-            let curBlockRef = this.codeBlockList[index];
-            if(curBlockRef != null && curBlockRef.current != null) {
-                if(index == selectIndex) {
-                    curBlockRef.current.style.background = "yellow";
-                } else {
-                    curBlockRef.current.style.background = "";
-                }
-            }
-        }  
     }
      
     getCodeBlockWithIndex(index: number): React.RefObject<HTMLElement> {
@@ -903,7 +892,8 @@ class Markdown extends React.Component<{ text: string; style?: any; extraClassNa
             h4: (props) => HeaderRenderer(props, 4),
             h5: (props) => HeaderRenderer(props, 5),
             h6: (props) => HeaderRenderer(props, 6),
-            code: (props) => this.CodeRenderer(props, codeSelect),
+            code: (props) => CodeRenderer(props), 
+            pre: (props) => this.CodeBlockRenderer(props, codeSelect), 
         };
         return (
             <div className={cn("markdown content", this.props.extraClassName)} style={this.props.style}>
--- a/src/app/common/common.tsx
+++ b/src/app/common/common.tsx
@@ -11,6 +11,7 @@ import cn from "classnames";
 import { If } from "tsx-control-statements/components";
 import type { RemoteType } from "../../types/types";
 import ReactDOM from "react-dom";
+import { GlobalModel } from "../../model/model";
 
 import { ReactComponent as CheckIcon } from "../assets/icons/line/check.svg";
 import { ReactComponent as CopyIcon } from "../assets/icons/history/copy.svg";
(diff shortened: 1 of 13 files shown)


wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1934,6 +1934,105 @@ func doOpenAICompletion(cmd *sstore.CmdType, opts *sstore.OpenAIOptsType, prompt
...
+
+func updateAsstResponseAndWriteToUpdateBus(ctx context.Context, cmd *sstore.CmdType, pk *packet.OpenAICmdInfoChatMessage, messageID int) {
+	update, err := sstore.UpdateWithUpdateOpenAICmdInfoPacket(ctx, cmd.ScreenId, messageID, pk)
+	if err != nil {
+		log.Printf("Open AI Update packet err: %v\n", err)
+	}
+	sstore.MainBus.SendScreenUpdate(cmd.ScreenId, update)
+}
+
+func doOpenAICmdInfoCompletion(cmd *sstore.CmdType, clientId string, opts *sstore.OpenAIOptsType, prompt []packet.OpenAIPromptMessageType) {
```
MrStashley, line 1953: This is a bit ugly, needs cleanup

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -2057,11 +2156,23 @@ func OpenAICommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (sstor
 	if err != nil {
 		return nil, fmt.Errorf("openai error, cannot make dyn cmd")
 	}
+	prompt := []packet.OpenAIPromptMessageType{{Role: sstore.OpenAIRoleUser, Content: promptStr}}
+	if resolveBool(pk.Kwargs["cmdinfo"], false) {
```
MrStashley, line 2160: I think eventually we also send the current command as a kwarg, then we can do the prompt engineering

wavesrv/pkg/sstore/dbops.go
```diff
@@ -854,6 +854,18 @@ func GetCmdByScreenId(ctx context.Context, screenId string, lineId string) (*Cmd
...
+func UpdateWithAddNewOpenAICmdInfoPacket(ctx context.Context, screenId string, pk *packet.OpenAICmdInfoChatMessage) (*ModelUpdate, error) {
+	ScreenMemAddCmdInfoChatMessage(screenId, pk)
+	cmdInfoUpdate := ScreenMemGetCmdInfoChat(screenId).Messages
+	return &ModelUpdate{OpenAICmdInfoChat: cmdInfoUpdate}, nil
+}
+
+func UpdateWithUpdateOpenAICmdInfoPacket(ctx context.Context, screenId string, messageID int, pk *packet.OpenAICmdInfoChatMessage) (*ModelUpdate, error) {
+	ScreenMemUpdateCmdInfoChatMessage(screenId, messageID, pk)
+	cmdInfoUpdate := ScreenMemGetCmdInfoChat(screenId).Messages
+	return &ModelUpdate{OpenAICmdInfoChat: cmdInfoUpdate}, nil
```
MrStashley, line 866: Currently sending the entire chat history every time
This seems to work fine without any issues of latency, and I think we can probably safely have a max history length to avoid crazy long histories, but another optimization is we can send just the newest message that is actually being streamed and then have the client update its array. I'm wondering if this is needed and I'm leaning towards that it is
This method is super wasteful, but it is small amounts of data
I guess the web socket is a shared and valuable resource

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1934,6 +1934,105 @@ func doOpenAICompletion(cmd *sstore.CmdType, opts *sstore.OpenAIOptsType, prompt
...
+	asstMessagePk := &packet.OpenAICmdInfoChatMessage{IsAssistantResponse: true, AssistantResponse: asstOutputPk, MessageID: asstOutputMessageID}
+	writePacketToUpdateBus(ctx, cmd, asstMessagePk)
+
+	doneWaitingForPackets := false
+	for !doneWaitingForPackets {
+		select {
+		case <-time.After(OpenAIPacketTimeout):
+			// timeout reading from channel
+			hadError = true
+			log.Printf("We need to figure out what to do about these errors; error in doOpenAICmdInfoCompletion")
```
MrStashley, line 1998: currently no way to error handle
I'm thinking add an error in the chat message type and then check for it in the client

wavesrv/pkg/cmdrunner/cmdrunner.go
MrStashley: Addressed

src/app/common/common.tsx
```diff
@@ -784,14 +785,78 @@ function HeaderRenderer(props: any, hnum: number): any {
...
+        super(props);  
+        this.codeBlockList = [];
+        if(this.props.codeSelect) {
+            let model = GlobalModel;
+            let inputModel = model.inputModel; 
+            this.codeSelectMarkdownID = inputModel.registerCodeSelectMarkdown();   
+        }
+    }  
+
+   CodeRenderer(props: any, codeSelect: boolean): any {
```
MrStashley, line 803: I moved this inside the markdown class, is that okay?

The pattern is kind of off now, but seems like functionality wise no one else was depending on this

src/model/model.ts
```diff
@@ -1677,6 +1693,130 @@ class InputModel {
...
+    }
+
+    setCodeSelectSelectedCodeBlock(blockIndex:number, markdownID:number) { 
+        if(markdownID != this.codeSelectMarkdownID) {
+            return;
+        }
+        if(blockIndex >= 0 && blockIndex < this.codeSelectBlockRefArray.length) {
+            this.codeSelectDeselectCodeBlock(this.codeSelectSelectedIndex, markdownID);
+            this.codeSelectSelectedIndex = blockIndex;
+            this.codeSelectBlockRefArray[blockIndex].current.style.background = "yellow";
```
MrStashley, line 1750: This is where I'm changing the style of the selected code block

src/model/model.ts
```diff
@@ -1677,6 +1693,130 @@ class InputModel {
...
+        }
+        return rtn;
+    }
+    
+    codeSelectDeselectCodeBlock(blockIndex: number, markdownID: number) {
+        if(markdownID != this.codeSelectMarkdownID) {
+            return;
+        }
+        if(blockIndex >= 0 && blockIndex < this.codeSelectBlockRefArray.length) {
+            this.codeSelectBlockRefArray[blockIndex].current.style.background = "";
```
MrStashley, line 1739: and this is where I'm changing it back

wavesrv/pkg/cmdrunner/cmdrunner.go
MrStashley: Addressed

src/model/model.ts
```diff
@@ -1228,7 +1229,16 @@ function getDefaultHistoryQueryOpts(): HistoryQueryOpts {
 class InputModel {
     historyShow: OV<boolean> = mobx.observable.box(false);
     infoShow: OV<boolean> = mobx.observable.box(false);
+    aIChatShow: OV<boolean> = mobx.observable.box(false);
     cmdInputHeight: OV<number> = mobx.observable.box(0);
+    aiChatTextAreaRef: React.RefObject<HTMLTextAreaElement>;
+    codeSelectBlockRefArray: Array<React.RefObject<HTMLElement>>;
+    codeSelectMarkdownID: number;
```
MrStashley, line 1236: Currently markdown id makes it so that only the most recent query will be part of the selection. 
I did this initially because I figured the past queries wouldn't be relevant and I thought it had potential for a big memory load, but I think we might want to rethink this and just have every code block generated as part of this selection

src/app/common/common.tsx
```diff
@@ -784,14 +785,66 @@ function HeaderRenderer(props: any, hnum: number): any {
...
+        if(codeSelect) {
+            let model = GlobalModel;
+            let inputModel = model.inputModel;
+            curBlockIndex = inputModel.addCodeBlockToCodeSelect(blockRef);
+            clickHandler = (e: React.MouseEvent<HTMLElement>, blockIndex: number) => {
+                inputModel.setCodeSelectSelectedCodeBlock(blockIndex);
+            };
+        } else {
+            clickHandler = (e: React.MouseEvent<HTMLElement>, blockIndex: number) => {
+                navigator.clipboard.writeText(codeText);
```
MrStashley, line 822: I added a simple click to copy functionality for all code blocks in markdown. There is no ui currently. I think it might make sense to remove this for this pr, or I can put it in as a small nice to have. In the future we want proper "click to copy" button functionality.

src/model/model.ts
```diff
@@ -1677,6 +1694,145 @@ class InputModel {
...
+                    if(!elementIsInView) {
+                        this.aiChatWindowRef.current.scrollTop = elemBottom - (this.aiChatWindowRef.current.clientHeight / 3);                        
+                    }
+                }
+            }
+            this.setAIChatFocus();
+        }  
+    }
+    
+    codeSelectDecrementCodeBlock() {
```
MrStashley, line 1761: This part is a bit wierd. I add code blocks to the end of the data structure so that each one preserves an index value, but when the user is cycling through them, we want to look through them in backwards order of time added, so from array length -1 to index 0. so "increment code select" selects the code block one above, but it actually selects the code block stored in the data structure at current selected index - 1. I don't wanna create tech debt and have one of you guys have to suffer through trying to change this in a few months lol, so I should look into adding comments or changing this to be more readable

src/model/model.ts
```diff
@@ -1228,7 +1229,16 @@ function getDefaultHistoryQueryOpts(): HistoryQueryOpts {
 class InputModel {
     historyShow: OV<boolean> = mobx.observable.box(false);
     infoShow: OV<boolean> = mobx.observable.box(false);
+    aIChatShow: OV<boolean> = mobx.observable.box(false);
     cmdInputHeight: OV<number> = mobx.observable.box(0);
+    aiChatTextAreaRef: React.RefObject<HTMLTextAreaElement>;
+    aiChatWindowRef: React.RefObject<HTMLDivElement>;
+    codeSelectBlockRefArray: Array<React.RefObject<HTMLElement>>;
```
MrStashley, line 1236: I wonder how readable this is 
code select is the name for storing the code blocks to select and copy them, and all the chat related variables are stored as ai chat or ai assistant chat, or cmd info chat. Maybe I should call them all ai cmd info for future clarity

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -2057,11 +2179,40 @@ func OpenAICommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (sstor
 	if err != nil {
 		return nil, fmt.Errorf("openai error, cannot make dyn cmd")
 	}
+	if resolveBool(pk.Kwargs["cmdinfo"], false) {
```
MrStashley, line 2182: all of the cmd info / chat related functionality goes through the open ai command. I didn't add any new base commands Does that make sense? I think it does

wavesrv/pkg/sstore/dbops.go
```diff
@@ -854,6 +855,26 @@ func GetCmdByScreenId(ctx context.Context, screenId string, lineId string) (*Cmd
 	})
 }
 
+func UpdateWithClearOpenAICmdInfo(screenId string) (*ModelUpdate, error) {
```
MrStashley, line 858: adding a lot of new functions here. I tried to make the names as descriptive as possible, let me know if they seem unclear

wavesrv/pkg/sstore/memops.go
```diff
@@ -43,11 +44,94 @@ func isIndicatorGreater(i1 string, i2 string) bool {
...
+	if ScreenMemStore[screenId] == nil {
+		ScreenMemStore[screenId] = &ScreenMemState{}
+	}
+	if ScreenMemStore[screenId].AICmdInfoChat == nil {
+		ScreenMemInitCmdInfoChat(screenId)
+	}
+	return ScreenMemStore[screenId].AICmdInfoChat
+}
+
+func ScreenMemUpdateCmdInfoChatMessage(screenId string, messageID int, msg *packet.OpenAICmdInfoChatMessage) {
```
MrStashley, line 120: this uses the index of the chat message in the chat message array to update messages in order to enable streaming

the caller needs to know this and call ScreenMemGetCmdInfoMessageCount and use that as the message id when creating a message

src/model/model.ts
MrStashley: addressed, no more markdown id

wavesrv/pkg/sstore/memops.go
sawka: maybe instead of doing the log.Printf, have this func return an `error`.

wavesrv/pkg/sstore/memops.go
```diff
@@ -43,11 +44,94 @@ func isIndicatorGreater(i1 string, i2 string) bool {
...
+func ScreenMemGetCmdInfoChat(screenId string) *OpenAICmdInfoChatStore {
+	MemLock.Lock()
+	defer MemLock.Unlock()
+	if ScreenMemStore[screenId] == nil {
+		ScreenMemStore[screenId] = &ScreenMemState{}
+	}
+	if ScreenMemStore[screenId].AICmdInfoChat == nil {
+		ScreenMemInitCmdInfoChat(screenId)
+	}
+	return ScreenMemStore[screenId].AICmdInfoChat
```
sawka, line 117: I think you're going to need to make a deep copy here to avoid any weird concurrency issues.

MrStashley: addressed

wavesrv/pkg/sstore/memops.go
MrStashley: addressed

src/model/model.ts
MrStashley: addressed, I renamed the functions to make them a little more clear, and added a comment

Issue #204, opened 2023-12-29
Cannot create multiple remote connections with same canonical name and different ports


Elliot-Construct:
**Describe the bug**

I have encountered a bug when trying to use WaveTerm to connect to multiple machines on the same url with the same credentials, but different ports. For example, I want to connect to user@domain:22 and user@domain:82. However, WaveTerm does not allow me to create new remote connections that use the same canonical name but a different port. It gives me an error message saying “cannot create remote “user@domain”: duplicate canonical name “user@domain”: cannot create new remote”. This prevents me from accessing different machines that share the same url and credentials.

**To Reproduce**
1. try to connect to multiple machines on same url with same credentials, port 22 and port 82.
Error: cannot create remote "user@domain": duplicate canonical name "user@domain": cannot create new remote

**Expected behavior**

The ability to create new remote connections that use the same canonical name but a different port.

**Desktop (please complete the following information):**
 - OS: [e.g. MacOS/Linux, x64 or arm64] Linux - Ubuntu - WSL

sawka:
thanks for submitting.  as canonicalname is the key for remotes this isn't possible right now.  i think we can fix this by adding non-standard port numbers to the canonical name.

as a workaround for now, you can add an entry in your /etc/hosts file to alias domain and then you'll be able to add a new connection.

oneirocosm:
Closing since this has been fixed with the merge of pull request #209

oneirocosm:
as a note, this will not be available in the next release (should be 0.6.0), but it can be found in the main github branch until then

PR #209, opened 2024-01-02
Allow dots in alias and add port to canonical name

### LINKED ISSUE #202: "Error: /remote:new invalid alias format" when adding a new connection
**Describe the bug**
New User here.
When I create a remote connection, I fill in the forms and receive this error:
```
Error: /remote:new invalid alias format
```

**To Reproduce**
1. Go to connections
2. Add new connection
3. Fill in boxes

I used a private key, and a public key. Is ED not supported?

**Expected behavior**
A new connection is made using a private key to the host detailed in the setup.

**Screenshots**

CleanShot 2023-12-28 at 15 34 56@2x

**Desktop (please complete the following information):**
 - OS:      macOS 14.2.1 (23C71)
 - Version: Darwin 23.2.0
 - ssh:     OpenSSH_9.4p1, LibreSSL 3.3.6
 
**Additional context**
I don't have a host config in .ssh/config for 10.0.0.8 here is my global host config.

```
Host *
    controlmaster auto
    controlpath /Users/bdmorin/.ssh/.tmp/%C
    controlpersist yes
    forwardagent yes
    forwardx11 no
    IdentityAgent "~/.1password/agent.sock"
    serveraliveinterval 5
    UserKnownHostsFile /dev/null
    StrictHostKeyChecking no
    addkeystoagent yes
    sendenv LS_COLORS
```
I don't know if waveterm uses .ssh/config file.

oneirocosm:
This has two changes to ssh.

The first is a small change that resolves #202 by allowing dots to be used in the alias for a remote connection.

The second is a slightly larger change that adds ports to the end of the canonical name if the port isn't 22. This addresses #204 by tracking each user/host/port as its own remote. As a small note, the down migration for this will delete duplicate user@hostname remotes to resolve conflicts with stripping off the port numbers in the canonical name. This should not be an issue for almost all users.

### DIFF
--- a/wavesrv/db/migrations/000029_canonicalport.up.sql
+++ b/wavesrv/db/migrations/000029_canonicalport.up.sql
@@ -1,4 +1,3 @@
 UPDATE remote
-SET remotecanonicalname = case when json_extract(sshopts, '$.issudo') then 'sudo@' else '' end ||
-       remoteuser || '@' || remotehost || COALESCE(':' || json_extract(sshopts, '$.sshport'), '')
+SET remotecanonicalname = remotecanonicalname || COALESCE( ":" || json_extract(sshopts, '$.sshport'), "")
 WHERE json_extract(sshopts, '$.sshport') != 22;
--- a/wavesrv/db/migrations/000029_canonicalport.up.sql
+++ b/wavesrv/db/migrations/000029_canonicalport.up.sql
@@ -1,0 +1,4 @@
+UPDATE remote
+SET remotecanonicalname = case when json_extract(sshopts, '$.issudo') then 'sudo@' else '' end ||
+       remoteuser || '@' || remotehost || COALESCE(':' || json_extract(sshopts, '$.sshport'), '')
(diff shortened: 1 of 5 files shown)


wavesrv/db/migrations/000029_canonicalport.up.sql
```diff
@@ -0,0 +1,4 @@
+UPDATE remote
+SET remotecanonicalname = case when json_extract(sshopts, '$.issudo') then 'sudo@' else '' end ||
```
sawka, line 2: I think you should just assume that remotecanonicalname is correct in the DB.  Because you're already checking that sshport is not 22 (and implicitly not null).  So then you can just do:
```
SET remotecanonicalname = remotecanonicalname || ':' || json_extract(sshopts, '$.sshport')
```

sawka:
Looks good, but I think changing the up migration will make this a bit safer (the comment on the down migration was just me being interested in what was going on there -- had to look it up -- and nothing needs to change).

oneirocosm:
I've simplified the up migration and have made it possible to add backslashes in the username as well

PR #210, opened 2024-01-03
When client is disconnected change log to show last 50 lines of wavesrv.logz

adred:


src/electron/emain.ts
```diff
@@ -485,6 +485,30 @@ electron.ipcMain.on("reload-window", (event) => {
...
+        const logPath = path.join(getWaveHomeDir(), "wavesrv.log");
+        const lastLines = await readLastLinesOfFile(logPath, numberOfLines);
+        event.reply("last-logs", lastLines);
+    } catch (err) {
+        console.error("Error reading log file:", err);
+        event.reply("last-logs", "Error reading log file.");
+    }
+});
+
+function readLastLinesOfFile(filePath, lineCount) {
```
sawka, line 499: this is going to be a problem if the log file becomes very large as we don't want to read it all into memory.  reading exactly the "n" last lines of a file is a hard problem.  in this case it might be easier to just run `tail -n [lines] wavesrv.log` and capture the output.

sawka:
I think you had some sort of a rollback of src/app/common/modals/modals.tsx?  All of the changes to TabSwitcherModal undo some of the work of the last commits in the Cmd-P PR.

src/electron/emain.ts
adred: Addressed

adred:
> I think you had some sort of a rollback of src/app/common/modals/modals.tsx? All of the changes to TabSwitcherModal undo some of the work of the last commits in the Cmd-P PR.

Addressed

PR #212, opened 2024-01-05
Add a modal confirmation before installing WaveShell

adred:


src/app/common/common.tsx
```diff
@@ -99,23 +99,57 @@ class Toggle extends React.Component<{ checked: boolean; onChange: (value: boole
...
 > {
+    generatedId;
+    static idCounter = 0;
+
+    constructor(props) {
+        super(props);
+        this.state = {
+            checkedInternal: this.props.checked !== undefined ? this.props.checked : Boolean(this.props.defaultChecked),
+        };
+        this.generatedId = `checkbox-${Checkbox.idCounter++}`;
```
sawka, line 120: do you need this id?  it doesn't seem to be used

sawka:
I don't think the DB migration is needed since a nil map[string]bool will work just fine for read-only access in go and you're doing the proper nil check in cmdrunner when setting a new flag.

src/app/common/common.tsx
adred: Yes. In case multiple instances of Checkbox are rendered. It's used as `htmlFor` value.

adred:
> I don't think the DB migration is needed since a nil map[string]bool will work just fine for read-only access in go and you're doing the proper nil check in cmdrunner when setting a new flag.

I think we do? `confirmflags.showShellPrompt` has to be set to true by default otherwise the prompt won't show at all.

PR #218, opened 2024-01-08
SSH Configuration Import Alert Modal

oneirocosm:
Previously, pressing the ssh configuration button left it unclear if anything changed. This changes that by popping up an alert modal that explains the deletions, creations, and updates made as a part of the import. The modal does not pop up when directly using the CLI.

Also, as a part of this change, the summary message that displays has been updated to give more detail about the types of changes that were made.


sawka:
Overall looks good.  Just need to change the user facing messages to say "connection" instead of the internal name "remote".

oneirocosm:
updated to replace the word "remote" with "connection" in the user-facing modal

PR #231, opened 2024-01-16
Command to copy file from remote to local 

MrStashley:
use /copyfile to copy a file from remote to local

this now works fully and can copy from a remote to local, or from a local to remote, or even from remote to remote, as well as simply from local to local 

the user can specify any of the above in the command syntax


wavesrv/pkg/remote/remote.go
```diff
@@ -600,7 +601,12 @@ func (msh *MShellProc) GetRemoteRuntimeState() RemoteRuntimeState {
 	if vars["remoteuser"] == "root" || vars["sudo"] == "1" {
 		vars["isroot"] = "1"
 	}
-	state.RemoteVars = vars
+	varsCopy := make(map[string]string)
+	// deep copy so that concurrent calls don't collide on this data
+	for key, value := range vars {
```
MrStashley, line 606: This change has nothing to do with my pr, I found a concurrency bug

the state returned from this function gets sent to the update websocket in `NotifyRemoteUpdate` (remote.go:615)
and vars is pointing to mshell memory, so when it gets marshaled to send to the client in `RunUpdates` (scws.go:129) -> wsshell `WriteJson` (wsshell.go:75), there is a race condition where any other call to `GetRemoteRuntimeState` will cause the memory in msh.Remote.StateVars to be both written to and read at the same time, even though there is a lock

sawka:
a couple ideas:
* for local to local, maybe just implement it with a localToLocal method -- open both files and run an io.Copy?
* would be nice to figure out how to integrate this with numRunningCommands and the status indicators.

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1102,6 +1107,526 @@ func SidebarRemoveCommand(ctx context.Context, pk *scpacket.FeCommandPacketType)
...
+		return fmt.Sprintf("%.2f Megabytes", mbSize)
+	}
+	kbSize := float64(size) / float64(1024)
+	if kbSize > 1 {
+		return fmt.Sprintf("%.2f Kilobytes", kbSize)
+	}
+	return fmt.Sprintf("%v Bytes", size)
+}
+
+func writeCmdStatus(ctx context.Context, cmd *sstore.CmdType, startTime time.Time, exitSuccess bool, outputPos int64) {
```
MrStashley, line 1126: Here is where I update the status indicators
@esimkowitz @sawka 
does this look correctly integrated?

I'm seeing on the terminal that the green check mark, the x and the running indicator works, and I also see the command in the running commands tab

I see there's also a `StatusIndicatorLevel_Output` in memops.go
Is that meant to be used when there is output to the terminal? not sure if I'm using that one correctly

esimkowitz: The `StatusIndicatorLevel_Output` is supposed to be set when there is new data written to stdout or stderr since the last time the user visited a tab. The `StatusIndicatorLevel` values are reset when a tab becomes active and then increment as running commands make updates. Wherever you receive an update packet that is not "Done", that should probably trigger an update to display `StatusIndicatorLevel_Output`.

Also, you should call `IncrementNumRunningCmds` or `IncrementNumRunningCmds_Update` and pass it a `1` when the command starts executing and `-1` when the command is finished executing, whether success or failure. This will indicate that the Running Commands spinner for the tab/workspace indicator should show. This is separate from the spinner that shows for a specific command line.

src/model/model.ts
```diff
@@ -1102,6 +1102,14 @@ class ScreenLines {
         let origCmd = this.cmds[cmd.lineid];
         if (origCmd != null) {
             origCmd.setCmd(cmd);
+        } else {
```
MrStashley, line 1105: This code handles a race condition where sometimes a copy file local to local finishes so fast that update cmd comes over the websocket before the add line update is returned, so when it tries to update a given line id, the line isn't there yet 

what I do is if the line is not found, I async wait 100 ms then try again, and in practice after that time the add line update has been processed so the line is there after the wait

wavesrv/pkg/remote/remote.go
```diff
@@ -2064,8 +2073,6 @@ func (msh *MShellProc) handleCmdDonePacket(donePk *packet.CmdDonePacketType) {
 			// fall-through (nothing to do)
 		}
 	}
-
-	go pushNumRunningCmdsUpdate(&donePk.CK, -1)
```
MrStashley, line 2068: moved this functionality into UpdateCmdDoneInfo, which gets called above in this function

is this okay?

wavesrv/pkg/remote/remote.go
```diff
@@ -1886,7 +1896,6 @@ func RunCommand(ctx context.Context, rcOpts RunCommandOpts, runPacket *packet.Ru
 		RunPacket: runPacket,
 	})
 
-	go pushNumRunningCmdsUpdate(&runPacket.CK, 1)
```
MrStashley, line 1889: moved this functionality into AddLineForCmd, which gets called earlier in this function

Any potential issues / regressions here?

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -697,6 +702,8 @@ func EvalCommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (sstore.
 	newPk, rtnErr := EvalMetaCommand(ctxWithHistory, pk)
 	if rtnErr == nil {
 		update, rtnErr = HandleCommand(ctxWithHistory, newPk)
+	} else {
```
MrStashley, line 705: bug fix - I found a bug where if I ran `/copyfile ~/'` there would be a nil ptr dereference because metacmd parsing fails and newPk here ends up as nil. rtnErr isn't properly checked here, so I fixed this by adding a return if there is an err here

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -266,6 +270,7 @@ func init() {
 	registerCmdFn("markdownview", MarkdownViewCommand)
 
 	registerCmdFn("csvview", CSVViewCommand)
+
```
MrStashley, line 273: should probably remove this whitespace

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1102,6 +1108,551 @@ func SidebarRemoveCommand(ctx context.Context, pk *scpacket.FeCommandPacketType)
...
+	var outputPos int64
+	outputStr := fmt.Sprintf("Copying [%v]:%v to [%v]:%v\r\n", sourceRemoteId.DisplayName, sourceFullPath, destRemoteId.DisplayName, destFullPath)
+	termopts := sstore.TermOpts{Rows: shellutil.DefaultTermRows, Cols: shellutil.DefaultTermCols, FlexRows: true, MaxPtySize: remote.DefaultMaxPtySize}
+	cmd, err := makeDynCmd(ctx, "copy file", ids, pk.GetRawStr(), termopts)
+	writeStringToPty(ctx, cmd, outputStr, &outputPos)
+	if err != nil {
+		// TODO tricky error since the command was a success, but we can't show the output
+		return nil, err
+	}
+	update, err := addLineForCmd(ctx, "/copy file", false, ids, cmd, "", nil)
```
MrStashley, line 1618: been looking into renderers for the file view task - is having "renderer" set to "" the right thing here?

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1102,6 +1108,551 @@ func SidebarRemoveCommand(ctx context.Context, pk *scpacket.FeCommandPacketType)
...
+	var exitSuccess bool
+	startTime := time.Now()
+	defer func() {
+		writeCmdStatus(ctx, cmd, startTime, exitSuccess, outputPos)
+	}()
+	localFile, err := os.Open(localPath)
+	if err != nil {
+		writeStringToPty(ctx, cmd, fmt.Sprintf("Error, unable to open file %v: %v\r\n", localFile, localPath), &outputPos)
+		return
+	}
```
sawka, line 1195: need a `defer localFile.Close()` here

wavesrv/pkg/cmdrunner/cmdrunner.go
sawka: ya, an empty string renderer defaults to "terminal"

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1102,6 +1108,551 @@ func SidebarRemoveCommand(ctx context.Context, pk *scpacket.FeCommandPacketType)
...
+	sstore.MainBus.SendScreenUpdate(cmd.ScreenId, update)
+	err = sstore.SetStatusIndicatorLevel(ctx, cmd.ScreenId, sstore.StatusIndicatorLevel_Output, false)
+	if err != nil {
+		// This is not a fatal error, so just log it
+		log.Printf("error setting status indicator level to output in writeStringToPty: %v\n", err)
+	}
+}
+
+func parseCopyFileParam(info string) (remote string, path string) {
```
sawka, line 1507: better convention here would be to return (string, string, error).  the error case can then return "", "", fmt.Errorf(...).  that feels more go-like than returning the remote name of "error"

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1102,6 +1108,551 @@ func SidebarRemoveCommand(ctx context.Context, pk *scpacket.FeCommandPacketType)
...
+	sourceRemote, sourcePath := parseCopyFileParam(sourceInfo)
+	var sourceRemoteId *ResolvedRemote
+	var destRemoteId *ResolvedRemote
+	if sourceRemote == "error" {
+		return nil, fmt.Errorf("error: malformed arguments - usage: [remote]:path ")
+	} else if sourceRemote == "" {
+		// use cur remote
+		sourceRemote = "connected"
+		sourceRemoteId = ids.Remote
+		if ids.Remote.DisplayName == "local" {
```
sawka, line 1538: this idea of whether the the remote is the "local" remote or not feels pretty fundamental.  let's move this check into sstore.go (close to the definition of RemoteType).  something like `func (r RemoteType) IsLocal() bool`.  I think the right check is `r.Local && !r.IsSudo()`

wavesrv/pkg/cmdrunner/cmdrunner.go
MrStashley: addressed

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1102,6 +1108,551 @@ func SidebarRemoveCommand(ctx context.Context, pk *scpacket.FeCommandPacketType)
...
+	if len(pk.Args) == 0 {
+		return nil, fmt.Errorf("usage: /copyfile [file to copy] local=[path to copy to on local]")
+	}
+	ids, err := resolveUiIds(ctx, pk, R_Screen|R_Session|R_RemoteConnected)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve connected remote id: %v", err)
+	}
+	sourceInfo := pk.Args[0]
+	sourceRemote, sourcePath := parseCopyFileParam(sourceInfo)
+	var sourceRemoteId *ResolvedRemote
```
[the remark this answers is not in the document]
MrStashley: addressed - I put them in resolver.go next the R_Remote, R_RemoteConnected ones

wavesrv/pkg/cmdrunner/cmdrunner.go
MrStashley: addressed

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1102,6 +1108,553 @@ func SidebarRemoveCommand(ctx context.Context, pk *scpacket.FeCommandPacketType)
...
+		return fmt.Sprintf("%.2f Megabytes", mbSize)
+	}
+	kbSize := float64(size) / float64(1024)
+	if kbSize > 1 {
+		return fmt.Sprintf("%.2f Kilobytes", kbSize)
+	}
+	return fmt.Sprintf("%v Bytes", size)
+}
+
+// this can only be called in a defer func, because recover() only works inside of a defe
```
MrStashley, line 1127: added comment for this function call and added defer prefix

PR #234, opened 2024-01-17
Intenational key binding fixes

MrStashley:
This PR creates a new function in the global model for checking key presses, and it creates a special key press syntax

An example of this syntax: if you want to check for ctrl + shift + e, you would write "ctrl:shift:e"

We also now use event.key instead of event.code for checking key presses 
There are a lot of considerations here
Event.key changes depending on modifiers - if user presses shift + e, event.key = "E" and if user pre sses option + e, on Mac event.key = "´"

I have handled the shift case, basically any of the above work - "Shift:e" == "Shift:E" == "E"
The option case however is still around, not sure how to go about handling that one
For now option keys aren't really going to work 
I guess the way is to go by event code rather than key, but then people have to write their option commands with different syntax, ie "Alt:KeyR" rather than "Alt:r"
We could also do some string conversions on the event code, like we can map "KeyR" to "r", and a lot of the non character keys like "Escape" or "Enter" or "ArrowDown" are the same both key and code. There will be a few special cases though, like "[" = "LeftBracket"

or we can just roll with the punches and make users do "Alt:®" for Alt + r (this is how it is working right now), or maybe there's a tool out there for finding the un optioned version of an option key

### DIFF
--- a/src/app/history/history.tsx
+++ b/src/app/history/history.tsx
@@ -95,14 +95,14 @@ function formatSessionName(snames: Record<string, string>, sessionId: string): s
 }
 
 @mobxReact.observer
-class HistoryCheckbox extends React.Component<{ checked: boolean, partialCheck?: boolean, onClick?: () => void }, {}> {
+class HistoryCheckbox extends React.Component<{ checked: boolean; partialCheck?: boolean; onClick?: () => void }, {}> {
     @boundMethod
     clickHandler(): void {
         if (this.props.onClick) {
             this.props.onClick();
         }
     }
-    
+
     render() {
         if (this.props.checked) {
             return <CheckedCheckbox onClick={this.clickHandler} className="history-checkbox checkbox-icon" />;
@@ -110,14 +110,18 @@ class HistoryCheckbox extends React.Component<{ checked: boolean, partialCheck?:
         if (this.props.partialCheck) {
             return (
                 <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="none">
-                    <rect x="0.5" y="0.5" width="15" height="15" rx="3.5" fill="#D5FEAF" fill-opacity="0.026"/>
-                    <path fill-rule="evenodd" clip-rule="evenodd" d="M4 8C4 6.89543 4.89543 6 6 6H10C11.1046 6 12 6.89543 12 8C12 9.10457 11.1046 10 10 10H6C4.89543 10 4 9.10457 4 8Z" fill="#58C142"/>
-                    <rect x="0.5" y="0.5" width="15" height="15" rx="3.5" stroke="#3B3F3A"/>
+                    <rect x="0.5" y="0.5" width="15" height="15" rx="3.5" fill="#D5FEAF" fill-opacity="0.026" />
+                    <path
+                        fill-rule="evenodd"
+                        clip-rule="evenodd"
+                        d="M4 8C4 6.89543 4.89543 6 6 6H10C11.1046 6 12 6.89543 12 8C12 9.10457 11.1046 10 10 10H6C4.89543 10 4 9.10457 4 8Z"
+                        fill="#58C142"
+                    />
+                    <rect x="0.5" y="0.5" width="15" height="15" rx="3.5" stroke="#3B3F3A" />
                 </svg>
-            );            
-        }
-        else {
-            return <div onClick={this.clickHandler} className="history-checkbox state-unchecked"/>
+            );
+        } else {
+            return <div onClick={this.clickHandler} className="history-checkbox state-unchecked" />;
         }
     }
 }
@@ -206,7 +210,7 @@ class HistoryView extends React.Component<{}, {}> {
     }
 
     @boundMethod
-    searchKeyDown(e: any) { 
+    searchKeyDown(e: any) {
         if (GlobalModel.checkKeyPressed(e, "Enter")) {
             e.preventDefault();
             GlobalModel.historyViewModel.submitSearch();
@@ -452,8 +456,8 @@ class HistoryView extends React.Component<{}, {}> {
                                 <div onClick={this.toggleSessionDropdown}>
                                     <span className="label">
                                         {hvm.searchSessionId.get() == null
-                                        ? "Limit Workspace"
-                                        : formatSessionName(snames, hvm.searchSessionId.get())}
+                                            ? "Limit Workspace"
+                                            : formatSessionName(snames, hvm.searchSessionId.get())}
                                     </span>
                                     <AngleDownIcon className="icon" />
                                 </div>
@@ -486,8 +490,8 @@ class HistoryView extends React.Component<{}, {}> {
                                 <div onClick={this.toggleRemoteDropdown}>
                                     <span className="label">
                                         {hvm.searchRemoteId.get() == null
-                                        ? "Limit Remote"
-                                        : formatRemoteName(rnames, { remoteid: hvm.searchRemoteId.get() })}
+                                            ? "Limit Remote"
+                                            : formatRemoteName(rnames, { remoteid: hvm.searchRemoteId.get() })}
                                     </span>
                                     <AngleDownIcon className="icon" />
                                 </div>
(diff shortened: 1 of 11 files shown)


src/app/workspace/cmdinput/aichat.tsx
```diff
@@ -76,8 +76,9 @@ class AIChat extends React.Component<{}, {}> {
             let inputModel = model.inputModel;
             let ctrlMod = e.getModifierState("Control") || e.getModifierState("Meta") || e.getModifierState("Shift");
             let resetCodeSelect = !ctrlMod;
+            console.log("Aichat event: " + e.key)
```
MrStashley, line 79: remove log

src/app/workspace/cmdinput/textareainput.tsx
```diff
@@ -232,59 +232,59 @@ class TextAreaInput extends React.Component<{ screen: Screen; onHeightChange: ()
                 }
                 inputModel.closeAIAssistantChat();
                 return;
-            }
-            if (e.code == "KeyE" && e.getModifierState("Meta")) {
+            }
```
MrStashley, line 235: should I call this meta instead of command?

src/model/model.ts
```diff
@@ -799,42 +809,45 @@ class Screen {
...
+            GlobalModel.checkKeyPressed(e, "PageUp") ||
+            GlobalModel.checkKeyPressed(e, "PageDown")) {
             return true;
         }
         return false;
     }
 
     termCustomKeyHandler(e: any, termWrap: TermWrap): boolean {
-        if (e.type == "keypress" && e.code == "KeyC" && e.shiftKey && e.ctrlKey) {
+        if (e.type == "keypress" && GlobalModel.checkKeyPressed(e, GlobalModel.KeyPressModifierShift + ":" + GlobalModel.KeyPressModifierControl + ":" + "c")) {
```
MrStashley, line 842: This is wrong, need to work on the "description". description needs to be able to handle and and or

src/model/model.ts
```diff
@@ -3544,6 +3563,68 @@ class Model {
         }
         return ProdServerEndpoint;
     }
+    
+    parseKeyDescription(description: string): Error | KeyPressType {
```
MrStashley, line 3567: unfortunately need to do this differently to support multiple mod keys, like ctrl + shift + c

I think I will use | for or, so ctrl + shift + c = "ctrl:shift:c"
but if we want either ctrl or shift and c, it will be ctrl|shift:c, and you could even do ctrl|opt|cmd:shift:c, etc

Then I think I will parse to an array, so ctrl:shift:c = ["ctrl", "shift", "c"], and ctrl|shift|cmd:c = [[ctrl, shift, cmd], "c"]

I think this makes sense, it's def a little complex tho

MrStashley: This is now implemented this way

sawka:
I think your CopyFile code got merged into this PR somehow?

src/app/workspace/cmdinput/textareainput.tsx
```diff
@@ -339,30 +340,36 @@ class TextAreaInput extends React.Component<{ screen: Screen; onHeightChange: ()
...
         }
-        if (e.code == "KeyC" && e.getModifierState("Control")) {
+        if (GlobalModel.checkKeyPressed(e, GlobalModel.KeyPressModifierControl + ":c")) {
             e.preventDefault();
             inputModel.resetInput();
             return;
         }
         if (
-            e.code == "KeyR" &&
-            (e.getModifierState("Meta") || e.getModifierState("Control")) &&
```
MrStashley, line 364: some or key bindings here
I feel like we should make these either Ctrl or Cmd, rather than both

src/app/workspace/cmdinput/aichat.tsx
MrStashley: addressed

src/model/model.ts
MrStashley: this has changed

src/app/history/history.tsx
```diff
@@ -95,29 +96,33 @@ function formatSessionName(snames: Record<string, string>, sessionId: string): s
 }
 
 @mobxReact.observer
-class HistoryCheckbox extends React.Component<{ checked: boolean, partialCheck?: boolean, onClick?: () => void }, {}> {
+class HistoryCheckbox extends React.Component<{ checked: boolean; partialCheck?: boolean; onClick?: () => void }, {}> {
```
MrStashley, line 99: prettier related changes, not rebase artifacts I'm pretty sure

src/app/workspace/cmdinput/aichat.tsx
```diff
@@ -91,17 +92,18 @@ class AIChat extends React.Component<{}, {}> {
...
                 }
             }
-            if (e.code == "Escape") {
+            if (checkKeyPressed(waveEvent, "Escape")) {
                 e.preventDefault();
                 e.stopPropagation();
                 inputModel.closeAIAssistantChat();
             }
-            if (e.code == "KeyL" && e.getModifierState("Control")) {
+
```
MrStashley, line 100: maybe remove space here
Idk what is your thought on spacing?
Apple was super strict on spacing inside of functions for some reason, so I'm hardwired to get rid of them

src/app/workspace/cmdinput/textareainput.tsx
MrStashley: I think it should just be control?
Looking at this though I understand why or is used here since it's a really important command

src/util/keyutil.ts
```diff
@@ -0,0 +1,137 @@
...
+import * as electron from "electron";
+
+type KeyPressDecl = {
+    mods: Record<ModKeyStrs, boolean>;
+    key: string;
+};
+
+function parseKeyDescription(keyDescription: string): KeyPressDecl {
+    let rtn = { key: "", mods: {} } as KeyPressDecl;
+    rtn.mods.Cmd = false;
```
MrStashley, line 11: is it ugly to init this way? wasn't sure how to properly init a record type
seems I have to set each of the values otherwise they end up as none

src/util/keyutil.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+    if (keyPress.mods.Shift && !event.shift) {
+        return false;
+    }
+    if (keyPress.mods.Ctrl && !event.control) {
+        return false;
+    }
+    let eventKey = event.key;
+    let descKey = keyPress.key;
+    if (eventKey.length == 1 && /[A-Z]/.test(eventKey.charAt(0))) {
```
MrStashley, line 58: So this accepts shift:e, shift:E or E all as the same thing
it also accepts "Space" or " " to mean space
Thoughts? 
Also I'm wondering if I should move this logic to the adapt keyboard event functions so that WaveKeyboardEvent has a normalized key value

sawka:
For event.key, I think space bar actually comes back as " " and not "Space" .  https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#whitespace_keys

src/util/keyutil.ts
sawka: we should add unicode checks in here.  instead of checking for [A-Z] you can use \p{Lu} (unicode uppercase characters).  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape .  I think you also need to set the `v` flag on the regexp.

see also https://unicode.org/reports/tr18/#General_Category_Property

src/util/keyutil.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+    if (descKey != eventKey) {
+        return false;
+    }
+    return true;
+}
+
+type ModKeyStrs = "Cmd" | "Option" | "Shift" | "Ctrl";
+
+interface WaveKeyboardEvent {
```
sawka, line 83: maybe cut down the number of lines of comments here.  since they are single line, just use `//`

sawka:
I think for WaveKeyboardEvent, it shouldn't have `cmd`.  You should just set `alt` and `meta` from the keyboard events so WaveKeyboardEvent always matches the native event.

put the special `cmd` logic in the check function.  also better since that logic will then only live in once place instead of multiple.

PR #241, opened 2024-01-22
Fixed chat bug "r.replace is not a function"

MrStashley:
This pr fixes the chat bug when rendering code blocks. The issue was that I had old code for finding the text value of a `<code>` block, but after a refactor, only `<pre>` elements were going through the CodeBlockRender function

### DIFF
--- a/src/app/common/common.tsx
+++ b/src/app/common/common.tsx
@@ -868,11 +868,11 @@ class Markdown extends React.Component<
     {}
 > {
     CodeBlockRenderer(props: any, codeSelect: boolean, codeSelectIndex: number): any {
-        let codeText:string ="";
+        let codeText: string = "";
         try {
             codeText = props.node.children[0].children[0].value;
-        } catch (e) { 
-            console.log("code block text not parsed correctly")
+        } catch (e) {
+            console.log("code block text not parsed correctly");
         }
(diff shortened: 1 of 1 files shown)


src/app/common/common.tsx
```diff
@@ -868,7 +868,12 @@ class Markdown extends React.Component<
     {}
 > {
     CodeBlockRenderer(props: any, codeSelect: boolean, codeSelectIndex: number): any {
-        let codeText = codeSelect ? props.node.children[0].children[0].value : props.children;
+        let codeText:string ="";
+        try {
+            codeText = props.node.children[0].children[0].value;
```
MrStashley, line 873: This should never fail, and it never has in practice, but I don't trust my understanding of the DOM well enough to be sure of this 
This try catch just makes sure it fails silently if any of the things I'm referencing don't exist, like node, or either of the children, or value 

is it okay to use try catch here?

MrStashley: Should I do some kind of error propagating? I think it's better if it fails silently if something bad happens. 
Because the markdown would still show, but the only thing that would break is the ability to copy the markdown
(note that in tracy's error case, this will no longer break, so copying will work as intended)

PR #259, opened 2024-01-29
set animation startTime to sync spinners for status indicators

sawka:


src/app/common/icons/icons.tsx
```diff
@@ -50,6 +55,35 @@ interface StatusIndicatorProps {
...
+    }
+
+    syncSpinner() {
+        if (!this.props.runningCommands) {
+            return;
+        }
+        if (this.iconRef.current == null) {
+            return;
+        }
+        let elem = this.iconRef.current;
```
esimkowitz, line 75:
```suggestion
        const elem = this.iconRef.current;
```

src/app/common/icons/icons.tsx
```diff
@@ -50,6 +55,35 @@ interface StatusIndicatorProps {
...
+        let spinElem = elem.querySelector(".spin");
```
esimkowitz, line 76:
```suggestion
        const spinElem = elem.querySelector(".spin");
```

src/app/common/icons/icons.tsx
```diff
@@ -50,6 +55,35 @@ interface StatusIndicatorProps {
...
+        if (spinElem == null) {
+            return;
+        }
+        let animArr = spinElem.getAnimations();
```
esimkowitz, line 80:
```suggestion
        const animArr = spinElem.getAnimations();
```

esimkowitz:
I am noticing that when switching tabs, the spinners can fall out of sync. Then, on some future update, they suddenly resume synchronization and flicker when they do so. Ideally, they should always be synchronized.

esimkowitz (approved):

PR #270, opened 2024-01-31
Shellpref for sshconfig import

oneirocosm:
The shell preference did not properly work with the imported ssh config options. This fixes that and makes it possible to manually edit the shell preference after the fact.


oneirocosm:
It might be worth adding a database migration to this. Just to ensure all previously imported remotes without a default shell preference get it automatically

PR #274, opened 2024-02-02
Fix term width in view connection detail modal

adred:
Screenshot 2024-02-02 at 11 32 33 PM


sawka:
The width of the terminal is also set on the backend at 80x25 (term opts passed to the PTY that we create).  Can we increase the width of the connections modals?  And then maybe put the terminal window into a scrollable container (this is necessary since users can increase the terminal font size).

adred:
@sawka  I've addressed this issue.

PR #277, opened 2024-02-07
FileView

MrStashley:
Todo for now: 

- download should pop up file viewer window, get file/ folder from pop up, and download to that location on local
- upload button that also pops up file viewer window

show download / upload progress on bottom status bar

- search functionality 
   - modal already gets search field
   - need to implement a searchdir / command on backend and going to need new mshell functionality 
   - this should use the go `filepath.walk` to recursively search a dir and return all matching files
   - maybe not a v1 thing, but I feel like the search modal lends itself to more complex search terms
      - file extension search
      - show hidden files check mark
      - regex
      - recursive / not recursive
     any other fields that we might want to show for file searching? 
     
  - click and drag functionality
  
  Should I break up this pr potentially in order to get file view in for testing? perhaps I could try to merge this after download / upload and then have search and click and drag be follow up prs? 

Issues: seems like cwd doesn't work correctly on remote machines, need to debug


src/plugins/plugins.ts
```diff
@@ -79,6 +80,16 @@ const PluginConfigs: RendererPluginType[] = [
         mimeTypes: ["image/*"],
         simpleComponent: SimpleImageRenderer,
     },
+    {
```
MrStashley, line 83: How does this look so far?

I am not exactly sure the difference between simple and full plugins, but I assume my plugin is simple because there probably won't be any packet streaming and it's probably less complex than codeedit

MrStashley: never mind, currently using full plugin

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -4221,6 +4243,72 @@ func MarkdownViewCommand(ctx context.Context, pk *scpacket.FeCommandPacketType)
...
+	update, err := addLineForCmd(ctx, "/"+GetCmdStr(pk), false, ids, cmd, "fileview", lineState)
+	if err != nil {
+		return nil, err
+	}
+	update.Interactive = pk.Interactive
+	sstore.MainBus.SendScreenUpdate(cmd.ScreenId, update)
+
+	go func() {
+		var outputPos int64
+		StatDir(context.Background(), ids, path, func(statPk *packet.FileStatPacketType, done bool, err error) {
```
MrStashley, line 4298: callback paradigm, thoughts? 
I thought this was a little less code than using a channel here 
and I think it works just as well because the callback should be a closure on the data that we need like the context and the cmd object etc

MrStashley: I guess the down side is the execution of the StatDir function stops while the callback is being executed
I think that is okay for this purpose because the callback is just gonna write to the pty and then return

but maybe a channel would make it faster

src/plugins/fileview/fileview.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    render() {
+        let model: FileViewRendererModel = this.props.model;
+        let dirList = model.dirList;
+        let file: any;
+        let index: number;
+        return (
+            <div>
+                <For each="file" index="index" of={dirList}>
+                    {this.renderFile(file, index)}
```
MrStashley, line 114: for now just listing a dir to test that it works

src/plugins/fileview/fileview.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+class FileViewRenderer extends React.Component<{ model: FileViewRendererModel }> {
+    renderFile(file: any, index: number) {
+        let keyString = "file-" + index;
+        return <div key={keyString}>{file.name}</div>;
+    }
+
+    render() {
+        let model: FileViewRendererModel = this.props.model;
+        let dirList = model.dirList;
```
MrStashley, line 108: how do I trigger the re render here?
Now I can see in the logs that all the dirs get sent almost immediately, but I have to trigger a redraw otherwise, for example by clicking on the line

MrStashley: Resolved

src/plugins/core/ptydata.ts
```diff
@@ -124,11 +117,13 @@ class PacketDataBuffer extends PtyDataBuffer {
             let ch = this.rawData[i];
             if (ch == NewLineCharCode) {
                 // line does *not* include the newline
-                let line = new TextDecoder().decode(
-                    new Uint8Array(this.rawData.buffer, this.parsePos, i - this.parsePos)
-                );
+                let curLineBytes = new Uint8Array(this.rawData.buffer, this.parsePos, i - this.parsePos);
+                let jsonStartIndex = curLineBytes.indexOf("{".charCodeAt(0));
+                let byteSize = curLineBytes.slice(jsonStartIndex, curLineBytes.length).length;
```
MrStashley, line 122: get size from raw byte array rather than later on from encoded rune string

src/plugins/fileview/fileview.tsx
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    @boundMethod
+    packetCallback(packetAny: any) {
+        let packet: T.FileInfoType = packetAny;
+        if (packet == null) {
+            return;
+        }
+        this.curDirectory = packet.path;
+        this.outputPos = packet.outputpos;
```
MrStashley, line 63: currently we store outputpos when we send commands, so that the new command can pick up where the old one left off. 

This honestly sucks and is gonna bring about a lot of timing issues as well as a lot of extra code, but I wasn't sure any other way to do this
Is there a way to get the current output pos from a file? 
I assume there are ways to get it with seek
I tried getting the file offset from the cirfile stat but that didn't work 

need to make a note to dig deeper into this and try and find a low level solution that doesn't involve sending output pos across the wire

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -4776,6 +4818,154 @@ func MarkdownViewCommand(ctx context.Context, pk *scpacket.FeCommandPacketType)
...
+func ViewDirCommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (sstore.UpdatePacket, error) {
+	ids, err := resolveUiIds(ctx, pk, R_Session|R_Screen|R_RemoteConnected)
+	if err != nil {
+		return nil, err
+	}
+	path := ids.Remote.FeState["cwd"]
+	if len(pk.Args) > 0 {
+		path = pk.Args[0]
+	}
+	cmd, err := sstore.GetCmdByScreenId(ctx, pk.Kwargs["screenid"], pk.Kwargs["lineid"])
```
MrStashley, line 4892: currently this doesn't conform with the "outputpty" paradigm. Need to implement a version of this that outputs to a line if "outputpty" is not specified, because I think this might be useful to a user as well

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1125,7 +1130,10 @@ func prettyPrintByteSize(size int64) string {
 }
 
 // this can only be called in a defer func, because recover() only works inside of a defe
-func deferWriteCmdStatus(ctx context.Context, cmd *sstore.CmdType, startTime time.Time, exitSuccess bool, outputPos int64) {
+func deferWriteCmdStatus(ctx context.Context, cmd *sstore.CmdType, startTime time.Time, exitSuccess bool, outputPos int64, outputPty bool) {
```
MrStashley, line 1133: if outputPty is true, no need to close the cmd, so this function just returns

I'm not sure if that is clear to someone who is reading this for the first time. I maybe should change the name of this variable or add a comment

yarn.lock
```diff
@@ -6616,6 +6616,11 @@ parseurl@~1.3.2, parseurl@~1.3.3:
   resolved "https://registry.yarnpkg.com/parseurl/-/parseurl-1.3.3.tgz#9da19e7bee8d12dff0513ed5b76957793bc2e8d4"
   integrity sha512-CiyeOxFT/JZyN5m0z9PfXw4SCBJ6Sygz1Dpl0wqjlhDEGGBP1GnsUVEL0p63hoG1fcj3fHynXi9NYO4nWOL+qQ==
 
+path-browserify@^1.0.1:
```
MrStashley, line 6619: Tried to get path to work in frontend js, but there ended up being more webpack issues that I could debug. For now, path.join works, but path.relative invokes a "process" package that webpack doesn't resolve correctly. I will need to get a path utility a different way, implement my own, or get some help here

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -1609,48 +1635,59 @@ func CopyFileCommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (sst
...
-			writeStringToPty(ctx, cmd, fmt.Sprintf("Couldn't connect to remote %v\r\n", sourceRemote), &outputPos)
+			writeStringToPty(ctx, cmd, fmt.Sprintf("Couldn't connect to remote %v\r\n", sourceRemote), &outputPos, outputPty)
 		} else {
-			writeStringToPty(ctx, cmd, "Auto connect successful\r\n", &outputPos)
+			writeStringToPty(ctx, cmd, "Auto connect successful\r\n", &outputPos, outputPty)
 		}
 	}
-	sstore.MainBus.SendScreenUpdate(cmd.ScreenId, update)
 	update = &sstore.ModelUpdate{}
 	if destRemote == LocalRemote && sourceRemote == LocalRemote {
```
MrStashley, line 1683: I'm thinking of refactoring this so that all copy files to "copyfileRemoteToRemote" and go through the mshell

This might be slower but it will remove a ton of boilerplate and make the code a lot more readable, and I think, since latency to the local wave shell is not currently a concern, it shouldn't impact speed

src/plugins/fileview/fileview.tsx
```diff
@@ -0,0 +1,523 @@
...
+                    "dir-container": file.isdir,
+                })}
+                key={keyString}
+                onClick={(event) => this.props.model.fileWasClicked(event, file)}
+            >
+                {file.name}
+                <If condition={!file.isdir}>
+                    <div className="download-button" onClick={(event) => downloadButtonClicked(event)}>
+                        <i className="fa-sharp fa-solid fa-download"></i>
+                        <input
```
MrStashley, line 401: Currently the file dialog on Mac says "upload". I couldn't find a way to change this online, does any one know how? 
It would be nice to be able to set a custom text like "upload" / "download", or just "open"

sawka:
merged main and fixed conflicts to try to keep this in sync so it doesn't drift too far.

PR #280, opened 2024-02-08
Refactor ModelUpdate to set up for decoupling sstore

esimkowitz:
This PR changes `ModelUpdate` mechanism from a statically-typed struct to an interface, allowing us to define the update mechanism and the update types separately. This sets us up to move app logic and update mechanisms into separate packages. Ultimately, `sstore` will only define low-level persistence logic.

With this new setup, `ModelUpdate` is an array of the ModelUpdateItem interface. Each update type will implement this interface and define its key. 

I've flattened some of the update types, such as session and screen, which previously were nested arrays. Now, for instance, each session update will be its own update object. 

I had to change how `Line` update works to fit this new array structure. Before, if both a Line update and a `Cmd` update were present in `ModelUpdate`, it was implicit that they be grouped together. This isn't possible with the new array structure, at least not without higher-level awareness of the update mechanism, which would get messy. Instead, I've added both `CmdType` and `LineType` as fields in the new `LineUpdate` type. This way, the relation between these two is more explicit.

The distinction between `Lines`, `Cmds`, and `ScreenLines` is fuzzy and should probably be reconsidered as we add more Line types.

I've also continued my crusade against lets, as well as a few other linting errors.

I've also renamed `model_old.ts` to `model_old.ts-deprecated` to make it clearer that we no longer use this. We should clean this up once we're sure we're done with it.

I've also fixed a small bug in the input component where a `mobx.action` wrapper was missing for a call to update an observable value.


wavesrv/pkg/sstore/updatebus.go
```diff
@@ -42,67 +41,70 @@ func (*PtyDataUpdate) UpdateType() string {
...
-	}
-	for idx, screen := range update.Screens {
-		if screen.ScreenId == newScreen.ScreenId {
-			update.Screens[idx] = newScreen
-			return
+func (update *ModelUpdate) append(item *ModelUpdateItem) {
+	*update = append(*update, item)
+}
+
+func AddUpdate[I ModelUpdateItem](update *ModelUpdate, item I) {
```
sawka, line 79: this doesn't need to be generic.  can just do `item ModelUpdateItem`.  could also make this variadic `item ...ModelUpdateItem`.

sawka: could also add a specific function for setting `Interactive`.  might make some of the code a bit simpler in cmdrunner.

PR #282, opened 2024-02-09
Declare types as global

adred:


src/app/appconst.ts
```diff
@@ -47,3 +47,13 @@ export const TabIcons = [
 export const VERSION = __WAVETERM_VERSION__;
 // @ts-ignore
 export const BUILD = __WAVETERM_BUILD__;
+
+/**
+ * Levels for the screen status indicator
+ */
+export enum StatusIndicatorLevel {
```
esimkowitz, line 54: Why does this need to go here? I'd prefer to keep it with the rest of the StatusIndicator code

adred: I was previously in types.ts. StatusIndicatorLevel can't be in custom.d.ts cos it's also used at runtime. I was hesitant to move it there because it's not a constant but behaves like a constant.

esimkowitz: it's fine, let's leave it in appconst

PR #283, opened 2024-02-09
removed i/l options in bash

panoskyriakis:
The interactive/login options in bash/zsh end up causing considerable lag when executing each command because they reload the .profile, .bashrc, .zshrc. Since the state is provided explicitly by the temporary rc files there is no need to reload the default files.


sawka:
can confirm that this does _significantly_ speed up command execution for zsh.  i want to run a couple more tests (and figure out why we were still running the slow zshrc files even when specifying ZDOTDIR to override them).  but looks promising!  thanks for submitting

PR #287, opened 2024-02-12
global shortcut for wave

sawka:
very simple global shortcut added to settings.  can set any Cmd-Fn key as the global shortcut (or Alt-Fn on linux/windows).  in the future we can support more complicated shortcuts.


src/electron/emain.ts
```diff
@@ -717,6 +747,7 @@ function runActiveTimer() {
     setTimeout(runActiveTimer, 5000); // start active timer, wait 5s just to be safe
     await app.whenReady();
     await createMainWindowWrap();
+    reregisterGlobalShortcut("Cmd+F4");
```
esimkowitz, line 750: How does this play on Linux? I think we should add an OS check here

esimkowitz: Also will this reset on start every time?

sawka: oops, i had two commits not pushed to github.  that Cmd+F4 was just testing (removed in the final).  I also have a check for linux which switches between Cmd (for MacOS), and Alt (for Linux/Windows).

sawka:
as a bonus, i also figured out why emain.ts had the strict null warnings (when every other file didn't).  it was excluded in tsconfig.json, removing that exclude has made typechecking on emain more consistent.

sawka:
merged main and fixed conflicts.  so, the shortcut comes from clientdata.  the FE re-registers the shortcut with the node process each time ClientData is updated.  the node process has code to no-op if it is unchanged.  this ensures we get the initial value set (when we first read clientdata) and pick up any updates.

src/app/clientsettings/clientsettings.tsx
```diff
@@ -13,6 +13,11 @@ import * as appconst from "@/app/appconst";
 
 import "./clientsettings.less";
 
+type DDItem = {
```
esimkowitz, line 16: What is `DDItem`? I think this could use a comment

src/app/clientsettings/clientsettings.tsx
```diff
@@ -105,6 +110,28 @@ class ClientSettingsView extends React.Component<{ model: RemotesModel }, { hove
         GlobalModel.clientSettingsViewModel.closeView();
     }
 
+    @boundMethod
+    handleChangeShortcut(newShortcut: string): void {
+        let prtn = GlobalCommandRunner.setGlobalShortcut(newShortcut);
```
esimkowitz, line 115: nit: const

src/app/clientsettings/clientsettings.tsx
```diff
@@ -67,10 +72,10 @@ class ClientSettingsView extends React.Component<{ model: RemotesModel }, { hove
         commandRtnHandler(prtn, this.errorMessage);
     }
 
-    getFontSizes(): any {
-        let availableFontSizes: { label: string; value: number }[] = [];
+    getFontSizes(): DDItem[] {
+        let availableFontSizes: DDItem[] = [];
```
esimkowitz, line 76: nit: const

src/app/clientsettings/clientsettings.tsx
```diff
@@ -105,6 +110,28 @@ class ClientSettingsView extends React.Component<{ model: RemotesModel }, { hove
...
     }
 
+    @boundMethod
+    handleChangeShortcut(newShortcut: string): void {
+        let prtn = GlobalCommandRunner.setGlobalShortcut(newShortcut);
+        commandRtnHandler(prtn, this.errorMessage);
+    }
+
+    getFKeys(): DDItem[] {
+        let opts: DDItem[] = [];
```
esimkowitz, line 120: nit: const

src/app/clientsettings/clientsettings.tsx
```diff
@@ -105,6 +110,28 @@ class ClientSettingsView extends React.Component<{ model: RemotesModel }, { hove
...
+        opts.push({ label: "Disabled", value: "" });
+        let platform = GlobalModel.getPlatform();
```
esimkowitz, line 122: nit: const

src/app/clientsettings/clientsettings.tsx
```diff
@@ -105,6 +110,28 @@ class ClientSettingsView extends React.Component<{ model: RemotesModel }, { hove
...
+        for (let i = 1; i <= 12; i++) {
+            let shortcut = (platform == "darwin" ? "Cmd" : "Alt") + "+F" + String(i);
```
esimkowitz, line 124: nit: const

src/app/clientsettings/clientsettings.tsx
```diff
@@ -105,6 +110,28 @@ class ClientSettingsView extends React.Component<{ model: RemotesModel }, { hove
...
+            opts.push({ label: shortcut, value: shortcut });
+        }
+        return opts;
+    }
+
+    getCurrentShortcut(): string {
+        let clientData = GlobalModel.clientData.get();
```
esimkowitz, line 131: nit: const

src/electron/emain.ts
```diff
@@ -698,6 +706,34 @@ function runActiveTimer() {
...
+    if (!waveutil.isBlank(currentGlobalShortcut)) {
+        if (electron.globalShortcut.isRegistered(currentGlobalShortcut)) {
+            electron.globalShortcut.unregister(currentGlobalShortcut);
+        }
+    }
+    if (waveutil.isBlank(shortcut)) {
+        currentGlobalShortcut = null;
+        return;
+    }
+    let ok = electron.globalShortcut.register(shortcut, () => {
```
esimkowitz, line 725: nit: const

esimkowitz (commented):
I think it's good to go, I added some nits, mostly variables that can be constants. The other was that `DDItem` is not an intuitive name for a type, I'm not really sure what it's meant to represent.

sawka:
fixed the type, and the consts

PR #290, opened 2024-02-13
Break model update code out of sstore

esimkowitz:
This establishes a new `feupdate` package, which replaces the primary functionality of the `updatebus.go` file in `sstore`. 

The recently-added user input mechanism has been moved to its own package `userinput`. 

These two packages share some core functionality, mainly the ability to register and unregister channels. This has been moved into its own package `updatebus` underneath `feupdate`. I decided it should be a sub-package since `feupdate` uses more of the base functionality.


oneirocosm:
This seems reasonable to me at a first glance. I'm going to take a closer look, but so far, it looks good to me.

PR #306, opened 2024-02-20
Persist line "minimized" status to DB

adred:
Note that I didn't change anything related to height logic so I don't see lines jumping when switching tabs or workspaces.


sawka:
small change, for this new LineState key lets use "wave:min" as opposed to "prompt:min".  can't change the old ones (since they are stuck in the DB), but going forward better to use the "wave:" prefix.

adred:
> small change, for this new LineState key lets use "wave:min" as opposed to "prompt:min". can't change the old ones (since they are stuck in the DB), but going forward better to use the "wave:" prefix.

addressed

PR #317, opened 2024-02-23
Wave 52

adred:


sawka:
Ah, we're going to have conflicts on the line styles.  I split lines.less into line.less and lines.less.  I've also replace a lot of the font declarations and terminal font sizes with CSS variables.  Most of that should merge, but you'll probably have to redo the lines.less work.

PR #322, opened 2024-02-23
Integrate SSH Library with Waveshell Installation/Auto-update

oneirocosm:
This fix is largely centered around restoring the old workflow where a failed connection allows waveshell to be automatically installed in specific cases. It also reorganizes a good amount of the code around creating a connection and keeping track of it.


waveshell/pkg/shexec/client.go
```diff
@@ -57,9 +57,26 @@ func (cw CmdWrap) Parser() (*packet.PacketParser, io.ReadCloser, io.ReadCloser,
 }
 
 func (cw CmdWrap) Start() error {
+	defer func() {
+		for _, extraFile := range cw.Cmd.ExtraFiles {
+			extraFile.Close()
```
sawka, line 62: when extraFile is nil, this technical works and just returns an error (https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/os/file_posix.go;l=19 -- does a nil check), but feels not so good.  add an `if` here and only call Close if extraFile is not nil.

Issue #284, opened 2024-02-12
"Markdown Preview" command (CMD +P) is duplicate with "switch to tab" shortcut.


gonzolively:
**Describe the bug**
When using `codeedit` to view markdown files, I issue a "CMD +P" command to use the markdown preview feature, but instead I get a "switch to tab" modal. Upon searching Wave shortcuts I do see "CMD +P" as the shortcut listed for tab switching. Consider switching the shortcut for one or the other.

**To Reproduce**
1. Type `codeedit` to view a file...ex "`codeedit README.md`
2. Issue `CMD + P` shortcut
3. Then presented with a "tab switcher" modal

**Expected behavior**
When I am using `codeedit` to view a file and issue the appropriate shortcut it should open a preview pane, instead I have to manually click that preview pane button as the "tab switcher" is what opens instead of the preview pane.

**Screenshots**
* [Screenrecording](https://github.com/wavetermdev/waveterm/assets/990275/7ac74bc0-51c7-4d24-bf03-d17caf9e0947)

**Desktop (please complete the following information):**
 - OS: MacOS Sanoma 14.3.1 (23D60)
 - Version: v0.6.1 (20240131-063625)

sawka:
Great catch.  Didn't see the conflict of the two keybindings.  (1) we shouldn't have these conflicting keybindings to begin with.  (2) the events are not processed correctly.  if the codeviewer is open it should have grabbed the Cmd-P first (overriding the global keybinding).

We're working on a better way to handle keybindings (and customize) them.

PR #340, opened 2024-02-27
Keybinding Manager

MrStashley:
This is a super rough draft so far, but here is an idea I had for how a key bind manager could be implemented 

Needs a lot of cleanup but I figured I would get some feedback before I go all in on refactor

I updated the electron keypress code to instead of sending specific key events, send just an "electron-key-press" event
Then the actual key events are processed in the browser process. 
The electron-key-press only gets sent when one of the keys that electron is using gets pressed, and then the global document keypress handler processes all other key events. I could just send electron-key-press on any key, but I figured we would want as little ipc communication as possible for performance

Actually on second thought I don't think we need the electron-key-press at all?
Have to think it through but I don't think we need the electron key press event when we can use use the document one instead 

There is a stack system where key presses get registered and when a key event comes (either from electron or from document keydown handler) the first item on the stack that matches the key press, ie the last one that was registered, gets a callback

Any part of the code can register key bindings. There is also a concept of "domain" which is just a unique string representing the current context. "domains" can get a domain handler that will be called when any of the keybindings registered to that domain are pressed. This can allow for more complex behavior. you can also register with the string "any" as your keybinding and you will get a callback if no other matches come up earlier in the stack 

This fixes issue 284 https://github.com/wavetermdev/waveterm/issues/284, because electron will register for "any" with domain "electron" before any lines are created, and then code edit will register for "Cmd:p" with domain "codeedit" and because of the stack paradigm, codeedit will get the callback because it registered second. when code edit closes it unregisters the domain "codeedit" removing all of its keybindings, and then the electron keybinding will work. 

Note: for now there can only be one "any" keybinding in the stack. I'm thinking I should implement a stopPropagation kind of thing, perhaps just look for a return value from the callback 

I think in the future this can be extended really nicely for user configured keybindings. We can have an array of key strings rather than descriptions, like "codeedit-close" and then have a mapping from string to keybinding and have a json of said mappings 

code edit could call GlobalModel.keybindManager.registerKeybinding("codeedit", "codeedit-close", ....) and we could do a quick convert from our map

We would just have to be careful about user input strings
I think there shouldn't be any issues there because a malformed keybinding just doesn't get called

And I think we should also have an alert for when keybindings change, because I feel like injecting files or data into waveterm is a potential attack vector


src/util/keyutil.ts
```diff
@@ -13,13 +13,119 @@ type KeyPressDecl = {
...
+    constructor() {
+        this.activeKeybinds = [];
+        this.domainCallbacks = new Map();
+    }
+}
+
 var PLATFORM: string;
 const PlatformMacOS: string = "darwin";
 
+function InitGlobalKeybindManager() {
```
MrStashley, line 112: TODO: remove

src/util/keyutil.ts
```diff
@@ -13,13 +13,119 @@ type KeyPressDecl = {
...
+
+function GetGlobalKeybindManager() {
+    return GlobalKeybindManager;
+}
+
 function setKeyUtilPlatform(platform: string) {
     PLATFORM = platform;
 }
 
+function keybindingIsEqual(bind1: string, bind2: string) {
```
MrStashley, line 125: TODO: this should compare to KeyPressDecls

src/electron/emain.ts
```diff
@@ -316,81 +322,71 @@ function createMainWindow(clientData: ClientDataType | null) {
...
-        if (input.type != "keyDown") {
+        if (waveEvent.type != "keyDown") {
             return;
         }
-        let mods = getMods(input);
+        let mods = getMods(waveEvent);
         if (checkKeyPressed(waveEvent, "Cmd:t")) {
-            win.webContents.send("t-cmd", mods);
-            e.preventDefault();
+            win.webContents.send("electron-key-press", mods);
```
MrStashley, line 330: checks if we are using this key and then sends "electron-key-press" event instead of individual event

src/models/model.ts
```diff
@@ -162,6 +174,69 @@ class Model {
             }
             return fontSize;
         });
+        getApi().onElectronKeyPress(this.onElectronKeyPress.bind(this));
+        this.keybindManager.registerKeybinding("electron", "any", (waveEvent) => {
```
MrStashley, line 178: This is where we handle electron key events. The reason why we move this logic over here is so that we can use the stack inside of keybindManager

src/models/model.ts
```diff
@@ -162,6 +174,69 @@ class Model {
...
+                this.onDigitCmd(digitNum, mods);
+            }
+            if (checkKeyPressed(waveEvent, "Cmd:[") || checkKeyPressed(waveEvent, "Cmd:]")) {
+                let rel = checkKeyPressed(waveEvent, "Cmd:]") ? 1 : -1;
+                this.onBracketCmd(rel, mods);
+                return;
+            }
+        });
         getApi().onTCmd(this.onTCmd.bind(this));
         getApi().onICmd(this.onICmd.bind(this));
```
MrStashley, line 241: all of these to be deleted

src/models/model.ts
```diff
@@ -477,7 +554,12 @@ class Model {
         return true;
     }
 
-    onWCmd(e: any, mods: KeyModsType) {
+    onElectronKeyPress(e: any, waveEvent: WaveKeyboardEvent) {
+        this.keybindManager.processKeyEvent(waveEvent);
```
MrStashley, line 558: pass event to key bind manager

src/models/model.ts
```diff
@@ -380,6 +455,8 @@ class Model {
 
     docKeyDownHandler(e: KeyboardEvent) {
         const waveEvent = adaptFromReactOrNativeKeyEvent(e);
+        console.log("got keydown");
+        this.keybindManager.processKeyEvent(waveEvent);
```
MrStashley, line 459: pass event to key bind manager

Stormfox2:
Is there a chance for this to be able to bind to keyIDs for people that switch between DVORAK and qwerty?

PR #385, opened 2024-03-06
SSH Quick Fixes

oneirocosm:
This updates a couple small things for the new release
- allows ssh usernames to start with numbers/capital letters and allows the use of capital letters in them in general
- use the ssh2 default identity file names
- allow a custom identity file to overwrite the config identity files instead of appending to them
- updates the styles of the user input modals so markdown requests look identical to non-markdown requests in terms of font styling

### DIFF
--- a/src/app/common/modals/userinput.less
+++ b/src/app/common/modals/userinput.less
@@ -11,3 +11,9 @@
         }
     }
 }
+
+.markdown {
+    font-family: var(--base-font);
+    font-size: 15px;
+    line-height: 1.5;
+}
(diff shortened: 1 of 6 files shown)


src/app/common/modals/userinput.less
```diff
@@ -11,3 +11,9 @@
         }
     }
 }
+
+.markdown {
+    font-family: var(--base-font);
```
sawka, line 16: updated how --base-font works.  now just write `font: var(--base-font)` and you get the font-family, size, weight, etc.

oneirocosm: good point--just pushed a fix

PR #390, opened 2024-03-06
fix: update versions in the work sum

oneirocosm:


oneirocosm:
closing since @sawka is working an another branch with similar changes

PR #404, opened 2024-03-07
Expose config path to frontend

MrStashley:
This pr adds an api endpoint for "api/config" which takes a path url parameter
It serves the file at WAVETERM_HOME/config/path with http.ServeFile

### DIFF
--- a/wavesrv/cmd/main-server.go
+++ b/wavesrv/cmd/main-server.go
@@ -680,17 +680,20 @@ func HandleRunCommand(w http.ResponseWriter, r *http.Request) {
 func AuthKeyMiddleWare(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		reqAuthKey := r.Header.Get("X-AuthKey")
+		w.Header().Set(CacheControlHeaderKey, CacheControlHeaderNoCache)
 		if reqAuthKey == "" {
+			log.Printf("mk2\n")
 			w.WriteHeader(500)
 			w.Write([]byte("no x-authkey header"))
 			return
 		}
 		if reqAuthKey != GlobalAuthKey {
+			log.Printf("mk1\n")
 			w.WriteHeader(500)
 			w.Write([]byte("x-authkey header is invalid"))
 			return
 		}
-		w.Header().Set(CacheControlHeaderKey, CacheControlHeaderNoCache)
+		log.Printf("mk3\n")
 		next.ServeHTTP(w, r)
 	})
 }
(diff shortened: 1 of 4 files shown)


wavesrv/cmd/main-server.go
```diff
@@ -618,6 +619,15 @@ func HandleReadFile(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func HandleConfig(w http.ResponseWriter, r *http.Request) {
+	qvals := r.URL.Query()
+	filePath := qvals.Get("path")
+	log.Printf("URL query: %v %v\n", qvals, filePath)
+	fullPath := path.Join(scbase.GetWaveHomeDir(), "config", filePath)
```
sawka, line 626: move this path definition to base.go (with the others).  "config" should be a const in there.

wavesrv/cmd/main-server.go
```diff
@@ -904,6 +914,7 @@ func main() {
 	gr.HandleFunc("/api/log-active-state", AuthKeyWrap(HandleLogActiveState))
 	gr.HandleFunc("/api/read-file", AuthKeyWrap(HandleReadFile))
 	gr.HandleFunc("/api/write-file", AuthKeyWrap(HandleWriteFile)).Methods("POST")
+	gr.HandleFunc("/api/config", AuthKeyWrap(HandleConfig))
```
sawka, line 917: this can be just `/config`

wavesrv/cmd/main-server.go
```diff
@@ -618,6 +619,15 @@ func HandleReadFile(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func HandleConfig(w http.ResponseWriter, r *http.Request) {
+	qvals := r.URL.Query()
+	filePath := qvals.Get("path")
```
sawka, line 624: don't pull the filePath off a query value, pull it directly off the URL!  so i can request `/config/keybindings.json`

wavesrv/cmd/main-server.go
```diff
@@ -618,6 +619,15 @@
...
+	log.Printf("URL query: %v %v\n", qvals, filePath)
+	fullPath := path.Join(scbase.GetWaveHomeDir(), "config", filePath)
+	log.Printf("fullPath: %v", fullPath)
+	http.ServeFile(w, r, fullPath)
```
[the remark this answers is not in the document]
MrStashley: good point 
does it really matter here? 
I think it doesn't, but if there's a quick way to not allow ".." might be an ok idea

I think we consider this to be open anyways, like if the user wanted to access the filesystem they can 
our threat model I think is more about stopping bad actors from executing code rather than limiting what they can do 

thoughts?

PR #432, opened 2024-03-12
App keybindings

MrStashley:
Adds the keybindings for switching main views as well as Cmd+P, and some refactor

Something to note: For the system keybinding for toggle dev tools, I've noticed that there is a small change from the expected behavior where if you are focusing the dev tools window you can't close the dev tools with the non accelerator keybinding. You can close the dev tools if you click on the wave window and unfocus dev tools. This is I think because when the dev tools is focused it captures all key bind events, and this is probably why it was originally an accelerator. Not a breaking issue, but something to think about. Also not directly related to this pr, just something I noticed and thought I would bring up

All of the keybindings in this pr seem to work exactly as expected

Also, the keybinding docs are slightly out of date. I can write up a quick pr for that one unless @gonzolively is working on it already


src/models/model.ts
```diff
@@ -719,8 +749,22 @@ class Model {
...
+        GlobalCommandRunner.bookmarksView();
+    }
+
+    onFocusCmdInputPressed() {
+        if (this.activeMainView.get() != "session") {
+            mobx.action(() => {
+                this.activeMainView.set("session");
+                setTimeout(() => {
+                    // allows for the session view to load
+                    this.inputModel.giveFocus();
```
MrStashley, line 762: I added a small timeout to allow for the session view to load before setting input focus in the case where session view is currently not active. Before it would switch to session view but not focus input

assets/default-keybindings.json
```diff
@@ -39,6 +39,14 @@
       "command": "app:openTabSearchModal",
       "keys": ["Cmd:p"]
     },
+    {
```
MrStashley, line 42: confirmed with custom keybindings that these work as intended

src/electron/emain.ts
```diff
@@ -377,11 +370,6 @@ function createMainWindow(clientData: ClientDataType | null): Electron.BrowserWi
             win.webContents.send("w-cmd", mods);
             return;
         }
-        if (checkKeyPressed(waveEvent, "Cmd:h")) {
-            win.webContents.send("h-cmd", mods);
-            e.preventDefault();
-            return;
-        }
         if (checkKeyPressed(waveEvent, "Cmd:p")) {
```
sawka, line 373: should this old cmd-p be removed?

sawka:
overall this looks good!  my only concern is that emain is still handling Cmd-I, Cmd-H, and Cmd-P?  Can you remove those bindings in emain, and the keybindManager handle those exclusively now, or is there some reason why you left them in emain?

src/electron/emain.ts
MrStashley: yeah, sorry about that, addressed

MrStashley:
> overall this looks good! my only concern is that emain is still handling Cmd-I, Cmd-H, and Cmd-P? Can you remove those bindings in emain, and the keybindManager handle those exclusively now, or is there some reason why you left them in emain?

Addressed Cmd-p
Cmd-h and Cmd-i are still being handled by emain as well? It seems to me like they are removed 

Ahh I see, apologies I forgot to remove the boilerplate for the ipc calls, but the ones for cmd-i and cmd-h were removed at the source so they weren't actually being called
Addressed

MrStashley:
Another little note - I reverified everything and I noticed that Cmd-i works while Cmd-p is active, which is probably not intended behavior
I think this isn't part of this pr, but we should probably treat Cmd-p as a modal in the future 

This should work fine with our react object keybindings model and our modal layer

PR #441, opened 2024-03-13
Slash Commands for keybindings

MrStashley:
First draft, I put in the keybindings that work out of the box with slash commands, and I am working on the others

Theoretically, multiple commands should work out of the box with this system, although I think the \n syntax might look a little wierd, maybe , separator is better

I couldn't get remote:show or client:show to work correctly. Maybe I'm using them incorrectly? It seems like remote:show didn't do anything when typed into the 0.7.0 prod release either

### DIFF
--- a/assets/default-keybindings.json
+++ b/assets/default-keybindings.json
@@ -41,7 +41,8 @@
     },
     {
       "command": "app:openConnectionsView", 
-      "keys": []
+      "keys": [],
+      "commandStr": "/remote:show"
     },
     {
       "command": "app:openSettingsView", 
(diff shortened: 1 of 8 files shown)


wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -3560,6 +3563,45 @@ func SessionSetCommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (s
 	return update, nil
 }
 
+func SleepCommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (scbus.UpdatePacket, error) {
```
MrStashley, line 3566: Not sure about this
For my slash command "scripts" it waits until the http call returns before running the next one, there is no update or anything, so for now at least, sleep has to lock the current process. I think this isn't too big a deal because we are already running inside of a goroutine? but either way I didn't end up using this so I can delete it if it seems sketchy

assets/default-keybindings.json
```diff
@@ -41,19 +41,22 @@
...
     },
     {
       "command": "app:newTab",
       "keys": ["Cmd:t"]
     },
     {
       "command": "app:focusCmdInput",
-      "keys": ["Cmd:i"]
+      "keys": ["Cmd:i"],
+      "commandStr":"/mainview session,/screen:set focus=cmd,/screen:set focus=input"
```
MrStashley, line 59: using `,` separator

MrStashley:
Should I remove the dev tools accelerator? 
Our toggle dev tools works, but it has the one quirk where you can't close it if you are focused on the dev tools, ie you have to click on the wave window with your mouse before the document key bindings work again

Issue #67, opened 2023-11-16
Add support for svg images to imageview command


tracymiranda:
Imageview supports many file types (png, gif, jpg) but not svgs currently

**Describe the solution you'd like**
imageview shows svg files inline in wave terminal

PR #453, opened 2024-03-14
Add support for svg images to imageview command

june23hy:
I've addressed issue #67!
Please take a look :)

### Description
I've modified the code to ensure that the MIME type is set to "image/svg+xml" exclusively for SVG files.
Below is how an SVG file appears when displayed using the imageview command.
image


esimkowitz:
It works great, thank you for this!! I'll merge it :)

PR #463, opened 2024-03-15
Main View keybindings

MrStashley:
I added all of the main view keybindings, including all of the session level ones

I feel like open side bar should automatically add the active line if a line is selected , thoughts? 

Also note: Cmd+ArrowDown does not currently work if a tab is selected, because the line terminal keybindings take over
Should I revert that one for now or just keep it the way it is? I feel like it's a small enough thing to keep in and I will work on the term ones next 
Those will be the hardest ones I think


src/app/workspace/workspaceview.tsx
```diff
@@ -17,6 +18,70 @@ import "./workspace.less";
...
+        });
+        keybindManager.registerKeybinding("pane", "session", "app:selectLineAbove", (waveEvent) => {
+            GlobalModel.onMetaArrowUp();
+            return true;
+        });
+        keybindManager.registerKeybinding("pane", "session", "app:selectLineBelow", (waveEvent) => {
+            GlobalModel.onMetaArrowDown();
+            return true;
+        });
+        keybindManager.registerKeybinding("pane", "session", "app:restartCommand", (waveEvent) => {
```
MrStashley, line 59: Does this organization make sense? for what is pane vs what is main view?

PR #482, opened 2024-03-20
Term keybindings

MrStashley:
This pr adds keybindings for the lines, including copy, and paste and arrow keys

I noticed that the normal copy and paste keys (Cmd+C and Cmd+V) seem to work fine in the terminal, I'm curious what are the Ctrl:Shift:C and Ctrl:Shift:V ones for?

It would be cool if there was a way to select a paragraph or something with keybindings but I think for now selection with keybindings is not possible? 

Either way I don't think they need to be removed or anything, just curious

### DIFF
--- a/src/models/screen.ts
+++ b/src/models/screen.ts
@@ -469,56 +469,12 @@ class Screen {
         this.renderers[lineId] = renderer;
     }
 
-    registerKeybindings(lineid: string) {
-        let keybindManager = GlobalModel.keybindManager;
-        let domain = "line-" + lineid;
-        let termWrap = this.terminals[lineid];
-        keybindManager.registerKeybinding("plugin", domain, "terminal:copy", (waveEvent) => {
-            let sel = termWrap.terminal.getSelection();
-            navigator.clipboard.writeText(sel);
-            return true;
-        });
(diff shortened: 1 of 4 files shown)


sawka:
Ah we add Ctrl:Shift:c and Ctrl:Shift:v because those are the standard terminal copy/paste keybindings.  (on windows it would be "Ctrl:c", but obviously that has meaning in the terminal, so most windows/linux users are used to Ctrl:Shift:c to do a copy in the terminal)

sawka:
not sure about this yet.  two things:
* I think we can move these keybindings to terminal.tsx (in the plugins directory).  and then play the same tricks we do elsewhere (mount/unmount).  then you can remove the keybindings from screen.ts
* haven't tested, but we need to make sure the terminal still captures all regular keystrokes and any keystroke with just the Ctrl modifier (e.g. Ctrl-C, Ctrl-A, Ctrl-E, etc.)

hooking into the focus handler was the right call.  but we need the TerminalRenderer (terminal.tsx) to also hook into the focus handler.  i can do that in `loadTerminal` when it calls screen.loadTerminalRenderer.  i think you need to hook into the focus handler there as well so that you get an OV<boolean> to change values inside of TerminalRenderer to trigger your keybindings on.

PR #483, opened 2024-03-20
Code edit keybindings

MrStashley:
Added code edit keybindings

There was some complexity to this one 

I found that if you click off of codeedit to deselect it, then press Cmd+D, the keybindings for "deleteActiveLine" gets called, which deletes the code edit line, instead of closing it
This is expected behavior and what should happen, but makes me wonder if we should make it so that if the user clicks off of the line window, ie if the user clicks on the sidebar etc, that we make it so there is no active line selected

### DIFF
--- a/src/plugins/code/code.tsx
+++ b/src/plugins/code/code.tsx
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import * as React from "react";
+import * as mobx from "mobx";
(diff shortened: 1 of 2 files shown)


sawka:
Good point.  It is almost like the line keybindings and the plugin keybindings always need to be active together.  So maybe for codeedit Cmd-D would always "close it".  and then once closed, it gives up its keybindigs so another Cmd-D would kill the line?

PR #484, opened 2024-03-20
Modal keybindings

MrStashley:
This currently doesn't work 

For some reason my ModalKeybindings class doesn't mount when a footer is on the screen, and I'm not sure why

### DIFF
--- a/src/app/common/elements/inlinesettingstextedit.tsx
+++ b/src/app/common/elements/inlinesettingstextedit.tsx
@@ -8,6 +8,8 @@ import { boundMethod } from "autobind-decorator";
 import cn from "classnames";
 import { If } from "tsx-control-statements/components";
 import { checkKeyPressed, adaptFromReactOrNativeKeyEvent } from "@/util/keyutil";
+import { GlobalModel } from "@/models";
+import { v4 as uuidv4 } from "uuid";
 
 import "./inlinesettingstextedit.less";
 
@@ -27,6 +29,11 @@ class InlineSettingsTextEdit extends React.Component<
     tempText: OV<string>;
     shouldFocus: boolean = false;
     inputRef: React.RefObject<any> = React.createRef();
+    curId: string;
+
+    componentDidMount(): void {
+        this.curId = uuidv4();
+    }
 
     componentDidUpdate(): void {
         if (this.shouldFocus) {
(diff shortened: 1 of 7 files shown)


src/app/common/elements/modal.tsx
```diff
@@ -32,8 +34,39 @@ interface ModalFooterProps {
...
+    }
+
+    render(): React.ReactNode {
+        return null;
+    }
+}
+
 const ModalFooter: React.FC<ModalFooterProps> = ({ onCancel, onOk, cancelLabel = "Cancel", okLabel = "Ok" }) => (
     <div className="wave-modal-footer">
+        <ModalKeybindings onOk={onOk} onCancel={onCancel}></ModalKeybindings>
```
MrStashley, line 69: modal keybindings class here

componentDidMount doesn't get called, even though it seems like this class should be on the screen

src/app/common/elements/modal.tsx
```diff
@@ -30,10 +33,53 @@ interface ModalFooterProps {
...
+
+const ModalFooter: React.FC<ModalFooterProps> = ({
+    onCancel,
+    onOk,
+    cancelLabel = "Cancel",
+    okLabel = "Ok",
+    keybindings = true,
+}) => (
     <div className="wave-modal-footer">
+        <If condition={keybindings}>
```
MrStashley, line 80: I kept keybindings as part of the footer but added an optional switch for it that's defaulted to true

The unusual footers use the class directly, but the modal footer has the keybindings built in still
Is this too magic? I figured it was convenient but I can get rid of it

src/app/common/elements/inlinesettingstextedit.tsx
```diff
@@ -60,24 +68,40 @@ class InlineSettingsTextEdit extends React.Component<
...
+        });
+    }
+
+    unregisterKeybindings() {
+        let domain = "inline-settings" + this.curId;
+        GlobalModel.keybindManager.unregisterDomain(domain);
+    }
+
+    handleBlur() {
+        console.log("handle blur");
```
sawka, line 97: remove console.log

src/app/common/elements/inlinesettingstextedit.tsx
```diff
@@ -60,24 +68,40 @@ class InlineSettingsTextEdit extends React.Component<
...
+        this.unregisterKeybindings();
+        this.cancelChange();
+    }
+
+    componentWillUnmount(): void {
+        console.log("handle unmoount");
```
sawka, line 103: move console.log

PR #487, opened 2024-03-21
Keybind UI utils

MrStashley:
This pr creates utils for displaying keybindings, including the ability to "pretty print" them, or show their respective keyboard image.

I also created an "info" section that can be added to a keybind list and can be queried. I think in the future this will allow us to get quick tool tips for each of the key binds, and also maybe in the future we could give chatGPT key bind info and allow it to give smart suggestions

### DIFF
--- a/src/models/model.ts
+++ b/src/models/model.ts
@@ -206,6 +206,7 @@ class Model {
             return resp.json();
         }).then((userKeybindings) => {
             this.keybindManager.setUserKeybindings(userKeybindings);
+            console.log("all keybindings:", this.keybindManager.getAllKeybinds());
         });
     }
 
--- a/src/models/model.ts
+++ b/src/models/model.ts
@@ -206,7 +206,6 @@ class Model {
             return resp.json();
         }).then((userKeybindings) => {
             this.keybindManager.setUserKeybindings(userKeybindings);
-            console.log("all keybindings:", this.keybindManager.getAllKeybindUIDescriptions());
         });
     }
 
--- a/src/models/model.ts
+++ b/src/models/model.ts
@@ -206,7 +206,7 @@ class Model {
             return resp.json();
         }).then((userKeybindings) => {
             this.keybindManager.setUserKeybindings(userKeybindings);
-            console.log("all keybindings:", this.keybindManager.getAllKeybinds());
+            console.log("all keybindings:", this.keybindManager.getAllKeybindUIDescriptions());
         });
     }
 
--- a/src/util/keyutil.ts
+++ b/src/util/keyutil.ts
@@ -488,10 +488,10 @@ function notMod(keyPressMod, eventMod) {
 
 function checkKeyPressed(event: WaveKeyboardEvent, keyDescription: string): boolean {
     let keyPress = parseKeyDescription(keyDescription);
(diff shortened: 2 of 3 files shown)


src/util/keyutil.ts
```diff
@@ -114,6 +123,79 @@ class KeybindManager {
         this.keyDescriptionsMap = newKeyDescriptions;
     }
 
+    prettyPrintKeybind(keyDescription: string): string {
+        let keyPress = parseKeyDescription(keyDescription);
+        let returnString = "";
+        if (keyPress.mods.Cmd) {
+            returnString += "⌘";
```
MrStashley, line 130: should I make these platform specific? I was wondering if I should make it the windows key on linux, idk

src/util/keyutil.ts
```diff
@@ -114,6 +123,79 @@ class KeybindManager {
...
+        if (keyPress.mods.Ctrl) {
+            returnString += "⌃";
+        }
+        if (keyPress.mods.Option) {
+            returnString += "⌥";
+        }
+        if (keyPress.mods.Shift) {
+            returnString += "⇧";
+        }
+        if (keyPress.mods.Meta) {
```
MrStashley, line 141: How the system currently works is Cmd is automatically MetaKey on Mac and AltKey on linux and option is the opposite. But if a Keybind specifies meta it's always looking for the meta key. kind of like how you can use key code to specify a physical key, I currently have it so meta and alt allow someone to override any platform specific behavior. Does that make sense? It's definitely a niche use case but I figured there's no reason not to have it

MrStashley: Also, in this case the other key would be ignored, so like if for some reason there is a keybinding that is `Meta:Cmd:c` that is interpreted as `Meta:c`, and `Alt:Option:c` is interpreted as `Option:c`. I feel like that is the most reasonable way to interpret those cases anyways

src/util/keyutil.ts
sawka: ya, that's a good idea to make them platform specific.

src/util/keyutil.ts
```diff
@@ -25,7 +25,8 @@ const KeyTypeCode = "code";
 
 type KeybindCallback = (event: WaveKeyboardEvent) => boolean;
 type KeybindConfigArray = Array<KeybindConfig>;
-type KeybindConfig = { command: string; keys: Array<string>; commandStr?: string };
+type KeybindConfig = { command: string; keys: Array<string>; commandStr?: string; info?: string };
+type KeybindUIDescription = { keys: Array<string>; info: string };
```
sawka, line 29: add "command" into this as well.  the command name carries a lot of information and is useful even without the info sometimes.

src/models/model.ts
```diff
@@ -206,6 +206,7 @@ class Model {
             return resp.json();
         }).then((userKeybindings) => {
             this.keybindManager.setUserKeybindings(userKeybindings);
+            console.log("all keybindings:", this.keybindManager.getAllKeybindUIDescriptions());
```
sawka, line 209: remove console.log.  we can see this by just accessing it directly in the console if we want it. `GlobalModel.keybindManager.getAllKeybindUIDescriptions()`

PR #489, opened 2024-03-22
Control keybindings

MrStashley:
Keybindings for the control level, which is basically all of the smaller things that could be on modals or on settings pages, etc

Tested and confirmed that inline settings  text edit works correctly on both the screen settings modal and the client settings main view page

Also, small thing that I found - the dropdown arrow keys don't scroll the dropdown. If you select an option that's scrolled out of reach, the selected item seems to disappear, unless you scroll with your mouse. 
it had been like that before I made these changes. I'm not exactly sure how to fix it, there wasn't a simple `scrollDropdown` lol, it seemed like it might involve some html stuff beyond my current knowledge. I can look into fixing it, but if anyone knows a simple way to do it feel free to let me know or push a commit

### DIFF
--- a/assets/default-keybindings.json
+++ b/assets/default-keybindings.json
@@ -24,12 +24,68 @@
       "keys": ["Backspace", "Delete"]
     },
     {
+      "command": "generic:space",
+      "keys": ["Space"]
+    },
+    {
+      "command": "generic:tab",
+      "keys": ["Tab"]
+    },
+    {
+      "command": "generic:numpad-0",
+      "keys": ["0"]
+    },
+    {
+      "command": "generic:numpad-1",
+      "keys": ["1"]
+    },
+    {
+      "command": "generic:numpad-2",
+      "keys": ["2"]
+    },
+    {
+      "command": "generic:numpad-3",
+      "keys": ["3"]
+    },
+    {
+      "command": "generic:numpad-4",
+      "keys": ["4"]
+    },
+    {
+      "command": "generic:numpad-5",
+      "keys": ["5"]
+    },
+    {
+      "command": "generic:numpad-6",
+      "keys": ["6"]
+    },
+    {
+      "command": "generic:numpad-7",
+      "keys": ["7"]
+    },
+    {
+      "command": "generic:numpad-8",
+      "keys": ["8"]
+    },
+    {
+      "command": "generic:numpad-9",
+      "keys": ["9"]
+    },
+    {
       "command": "generic:selectAbove",
       "keys": ["ArrowUp"]
     },
     {
       "command": "generic:selectBelow",
       "keys": ["ArrowDown"]
+    },
+    {
+      "command": "generic:selectLeft",
+      "keys": ["ArrowLeft"]
+    },
+    {
+      "command": "generic:selectRight",
+      "keys": ["ArrowRight"]
     },
     {
       "command": "generic:selectPageAbove",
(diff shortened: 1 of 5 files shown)


src/app/common/elements/datepicker.tsx
```diff
@@ -315,51 +319,74 @@ const DatePicker: React.FC<DatePickerProps> = ({ selectedDate, format = "MM/DD/Y
...
-        const key = event.key;
+    const handleKeyDown = (event, currentPart) => {};
 
-        if (key === "ArrowLeft" || key === "ArrowRight") {
-            // Handle arrow navigation without selecting text
-            handleArrowNavigation(event, currentPart);
-            return;
-        }
+    const handleFocus = (event, part) => {
+        event.target.select();
```
MrStashley, line 325: currently when you select the date picker for the first time, handleFocus gets called twice and so the keys get registered twice. I think that's because handleMouseDown calls handleFocus which calls event.target.select() which I think triggers handleFocus again ? 

Either way it's not a big deal and doesn't affect anything. I could in theory add an mobx observable here so that they don't register twice. like keybindsRegistered: OV<boolean> which then gets set to false when the keybinds get unregistered

I think this is worth doing because there's a log print that happens if a key bind gets registered twice

MrStashley: addressed - I implemented it as described

src/app/common/elements/datepicker.tsx
```diff
@@ -315,51 +327,91 @@ const DatePicker: React.FC<DatePickerProps> = ({ selectedDate, format = "MM/DD/Y
...
+        keybindManager.registerKeybinding("control", domain, "generic:space", (waveEvent) => {
             toggleModal();
-            return;
+            return true;
+        });
+        keybindManager.registerKeybinding("control", domain, "generic:confirm", (waveEvent) => {
+            toggleModal();
+            return true;
+        });
+        keybindManager.registerKeybinding("control", domain, "generic:cancel", (waveEvent) => {
```
MrStashley, line 362: I added enter, escape and tab keybindings because I felt like they would be nice. enter toggles modal, escape closes modal, and tab does the same thing as arrow right, selects the next part of the date

assets/default-keybindings.json
```diff
@@ -23,6 +23,54 @@
       "command": "generic:deleteItem",
       "keys": ["Backspace", "Delete"]
     },
+    {
+      "command": "generic:space",
+      "keys": ["Space"]
+    },
+    {
+      "command": "generic:tab",
```
MrStashley, line 31: some big picture thoughts, does it make sense to limit keybindings to keyDescriptions? I think for now it does

I guess for plugins, we would have the plugin include a keybindings file? I think that makes sense 
Like you could download a plugin and it would be like a folder zip with .tsx files and a keybindings.json or maybe a <any-prefix>keybindings.json 
And we would scan our plugins folder, find all of these keybindings files and integrate them into our keybindings map? 
And we would "force" plugins to declare their keybindings in this way

I think that makes sense

We can't really force plugins to use our keybindings api currently, if we let them give us react elements I don't know a way that we can stop them from using onKeyDown, but I think that's fine I don't think we're really concerned with malicious plugins, it's more that plugins would want to declare their keybindings through our api so that they show up in the command palette and people can rebind them

MrStashley: I also wonder if this is limiting for us in the future
I guess not, and I guess if we come across a future case where we really want to use a keybinding that doesn't make sense to just throw on the default-keybindings.json list, it is very easy to change our system to allow registering key descriptions directly. I don't think we will end up in a situation where we feel limited by this design choice, I think that the system is flexible enough

PR #492, opened 2024-03-26
Keybind chords

MrStashley:
Got key bind chords working

How they work for now, is each key press in the chord will also be evaluated, so if you press Ctrl:Shift:o and then "c", if there is a binding for Ctrl:Shift:o, that binding will be triggered, and then when you enter "c", if you have a chord binding for "Ctrl:Shift:o+c", then c will be evaluated, however, if there is a "c" binding that comes before in the stack, that key will be called first. 
That is something I have to think about. I feel like most of the last keys in a chord will be single keys, so I think it shouldn't matter too much if we also evaluate the single key, but we should think about if the correct behavior is that the single key is not evaluated. I'm not sure how to implement that from what I have but I'm sure it's possible


MrStashley:
Gonna close this for now, will potentially reopen it later

PR #494, opened 2024-03-26
User input keybindings fix

MrStashley:
Pretty straight forward, user input had keybindings but the Modal.Footer already has keybindings built in 

An argument could be made that the Modal.Footer keybindings are too magic, but in this case I think it would have been pretty hard to put the keybindings in the userInput, at least with my limited react knowledge

### DIFF
--- a/src/app/common/modals/userinput.tsx
+++ b/src/app/common/modals/userinput.tsx
@@ -1,9 +1,7 @@
 import * as React from "react";
-import * as mobx from "mobx";
 import { GlobalModel } from "@/models";
 import { Choose, When, If } from "tsx-control-statements/components";
 import { Modal, PasswordField, TextField, Markdown, Checkbox } from "@/elements";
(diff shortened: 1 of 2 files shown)


src/app/common/modals/userinput.tsx
```diff
@@ -44,22 +43,6 @@ export const UserInputModal = (userInputRequest: UserInputRequest) => {
         [userInputRequest]
     );
 
-    function handleTextFocus() {
```
MrStashley, line 47: For now, if the user is not focused on the user input modal, they can still remove it or confirm it with escape or enter

I think this is the correct behavior, because we wouldn't want the keybindings to unregister and then have the user be able to access the layer below

I briefly thought of disabling rather than unregistering the keybindings if they're not in focus, and I can do that in the future if we want, but I think for now it makes sense not to disable them, because the modal is the only thing on the screen anyway

PR #514, opened 2024-03-27
Add esc keybinding to all modals, add update button in About modal

esimkowitz:


esimkowitz (commented):
accidentally made in my fork, oops

PR #535, opened 2024-03-30
Replace React with Preact

esimkowitz:
This is a lift-and-shift replacement of React with Preact using the preact/compat layer to seamlessly transition over.


esimkowitz:
Abandoning for now

PR #538, opened 2024-04-01
Pull in linting fixes from preact branch

esimkowitz:
Decided not to pursue the Preact migration for now, instead I'm pulling in some linting fixes that I had made while experimenting with Preact

### DIFF
--- a/src/app/workspace/cmdinput/textareainput.tsx
+++ b/src/app/workspace/cmdinput/textareainput.tsx
@@ -667,7 +667,7 @@ class TextAreaInput extends React.Component<{ screen: Screen; onHeightChange: ()
                 <textarea
                     key="main"
                     ref={this.mainInputRef}
-                    spellCheck={false}
+                    spellCheck="false"
                     autoComplete="off"
                     autoCorrect="off"
                     id="main-cmd-input"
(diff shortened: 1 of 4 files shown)


src/app/workspace/cmdinput/textareainput.tsx
```diff
@@ -670,7 +667,7 @@ class TextAreaInput extends React.Component<{ screen: Screen; onHeightChange: ()
                 <textarea
                     key="main"
                     ref={this.mainInputRef}
-                    spellCheck="false"
```
sawka, line 673: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck .  In this case I think the string "false" is correct.  same for the input box a couple lines below.

src/app/workspace/screen/screenview.tsx
```diff
@@ -399,10 +403,11 @@ class ScreenWindowView extends React.Component<{ session: Session; screen: Scree
         super(props);
         this.setSize_debounced = debounce(1000, this.setSize.bind(this));
         this.windowViewRef = React.createRef();
+        this.props_ = props;
     }
 
     setSize(width: number, height: number): void {
-        let { screen } = this.props;
+        const { screen } = this.props_;
```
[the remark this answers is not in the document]
esimkowitz: When trying out Preact, mobx was complaining about the way we accessed the props here. Copying the props to an observable solved it, not sure why exactly.

src/app/workspace/cmdinput/textareainput.tsx
esimkowitz: The type accepted is Booleanish, which accepts bool | "true" | "false". Preact had it as only a bool, so I had to change it. I can revert though

PR #540, opened 2024-04-01
Remove log for duplicate key register

MrStashley:
This pr removes the log for registering duplicate keybindings. I think we should do this at some point, but open to more discussion on this one


oneirocosm:
As a temporary measure, would it be reasonable to have at active on dev builds only?

PR #541, opened 2024-04-01
Autocomplete for Wave

esimkowitz:
This adds command autocomplete, as well as revamped file and directory autocomplete suggestions. When typing a command with autocomplete enabled, ghost text at the end of the line will show the first suggestion returned by the suggestion runtime. Hitting tab twice will display all suggestions.


esimkowitz:
closing in favor of #599

PR #547, opened 2024-04-04
Use style tag to override terminal theme

adred:


src/app/common/elements/styleblock.tsx
```diff
@@ -0,0 +1,72 @@
...
+    loadThemeStyles() {
+        const { themeKey, termTheme, scope } = this.props;
+        const currTheme = termTheme[themeKey];
+
+        if (currTheme && currTheme !== this.theme && currTheme) {
+            const rtn = GlobalModel.getTermThemeJson(currTheme);
+            rtn.then((termThemeJson) => {
+                if (termThemeJson && typeof termThemeJson === "object") {
+                    const styleProperties = Object.entries(termThemeJson)
+                        .map(([key, value]) => `--term-${key}: ${value};`)
```
sawka, line 39: need some validation on key/value so they don't break the CSS.

adred: addressed

PR #573, opened 2024-04-12
Sudo Caching

oneirocosm:
This change allows waveterm to cache your sudo password for use across multiple ptys. This prevents the user from needing to enter it for every command where it's required.

### DIFF
--- a/wavesrv/pkg/cmdrunner/cmdrunner.go
+++ b/wavesrv/pkg/cmdrunner/cmdrunner.go
@@ -610,7 +610,6 @@ func RunCommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (scbus.Up
 	if err != nil {
 		return nil, fmt.Errorf("/run error, invalid lang: %w", err)
 	}
-	sudoArg := resolveBool(pk.Kwargs[KwArgSudo], false)
 
 	cmdStr := firstArg(pk)
 	expandedCmdStr, err := doCmdHistoryExpansion(ctx, ids, cmdStr)
@@ -643,7 +642,11 @@ func RunCommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (scbus.Up
 	}
 	runPacket.Command = strings.TrimSpace(cmdStr)
 	runPacket.ReturnState = resolveBool(pk.Kwargs["rtnstate"], isRtnStateCmd)
-	runPacket.IsSudo = sudoArg || IsSudoCommand(cmdStr)
+	if sudoArg, ok := pk.Kwargs[KwArgSudo]; ok {
+		runPacket.IsSudo = resolveBool(sudoArg, false)
+	} else {
+		runPacket.IsSudo = IsSudoCommand(cmdStr)
+	}
 	rcOpts := remote.RunCommandOpts{
 		SessionId:     ids.SessionId,
 		ScreenId:      ids.ScreenId,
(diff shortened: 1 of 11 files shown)


wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -3916,6 +3921,28 @@ func DebugRemoteInstanceCommand(ctx context.Context, pk *scpacket.FeCommandPacke
...
+func ClearSudoCache(ctx context.Context, pk *scpacket.FeCommandPacketType) (rtnUpdate scbus.UpdatePacket, rtnErr error) {
+	ids, err := resolveUiIds(ctx, pk, R_Session|R_Screen|R_Remote)
+	if err != nil {
+		return nil, err
+	}
+	ids.Remote.MShell.ClearCachedSudoPw()
+
+	clearAll := resolveBool(pk.Kwargs["all"], false)
+	if clearAll {
+		for _, proc := range remote.GlobalStore.Map {
```
sawka, line 3933: you can't iterate directly over GlobalStore.Map (it needs to be synchronized with GlobalStore.Lock).  instead call `remote.GetRemoteMap()` (which makes a copy of the map using the lock for protection).  you can then safely iterate over the copy.

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -3916,6 +3921,28 @@ func DebugRemoteInstanceCommand(ctx context.Context, pk *scpacket.FeCommandPacke
...
+			proc.ClearCachedSudoPw()
+		}
+	}
+
+	update := scbus.MakeUpdatePacket()
+	update.AddUpdate(sstore.InfoMsgType{
+		InfoMsg:   "sudo password cleared",
```
sawka, line 3940: if clearAll is specified, return a slightly different info message -- like "all sudo passwords cleared"

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -639,6 +643,7 @@ func RunCommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (scbus.Up
 	}
 	runPacket.Command = strings.TrimSpace(cmdStr)
 	runPacket.ReturnState = resolveBool(pk.Kwargs["rtnstate"], isRtnStateCmd)
+	runPacket.IsSudo = sudoArg || IsSudoCommand(cmdStr)
```
sawka, line 646: slightly different logic here. `if _, ok := pk.Kwargs[KwArgSudo]; ok` then use sudoArg otherwise use IsSudoCommand.  that lets the user use `[sudo=0]` to force IsSudo to be false.

wavesrv/pkg/remote/remote.go
```diff
@@ -164,7 +168,10 @@ type MShellProc struct {
 	RunningCmds      map[base.CommandKey]*RunCmdType
 	PendingStateCmds map[pendingStateKey]base.CommandKey // key=[remoteinstance name] (in progress commands that might update the state)
 
-	Client *ssh.Client
+	Client          *ssh.Client
+	sudoPw          *[]byte
```
sawka, line 172: sudoPw should be a `[]byte` not a `*[]byte`.  it is a slice, so it is still possible to set it to nil.

wavesrv/pkg/cmdrunner/cmdrunner.go
oneirocosm: fixed

wavesrv/pkg/cmdrunner/cmdrunner.go
oneirocosm: fixed

PR #575, opened 2024-04-15
auxview focus + keybindings

MrStashley:
A few fixes for focus and keybindings around auxview

### DIFF
--- a/src/app/workspace/cmdinput/aichat.tsx
+++ b/src/app/workspace/cmdinput/aichat.tsx
@@ -57,18 +57,13 @@ class AIChat extends React.Component<{}, {}> {
     chatListKeyCount: number = 0;
     chatWindowScrollRef: React.RefObject<HTMLDivElement>;
     textAreaRef: React.RefObject<HTMLTextAreaElement>;
-    isFocused: OV<boolean>;
     termFontSize: number = 14;
 
     constructor(props: any) {
         super(props);
         this.chatWindowScrollRef = React.createRef();
         this.textAreaRef = React.createRef();
-        this.isFocused = mobx.observable.box(false, {
-            name: "aichat-isfocused",
-        });
     }
-
     componentDidMount() {
         const inputModel = GlobalModel.inputModel;
         if (this.chatWindowScrollRef?.current != null) {
@@ -110,14 +105,12 @@ class AIChat extends React.Component<{}, {}> {
 
     onTextAreaFocused(e: any) {
         mobx.action(() => {
-            this.isFocused.set(true);
             GlobalModel.inputModel.setAuxViewFocus(true);
         })();
     }
 
     onTextAreaBlur(e: any) {
         mobx.action(() => {
-            this.isFocused.set(false);
             GlobalModel.inputModel.setAuxViewFocus(false);
         })();
     }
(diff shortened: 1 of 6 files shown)


src/app/workspace/cmdinput/aichat.tsx
```diff
@@ -110,12 +111,14 @@ class AIChat extends React.Component<{}, {}> {
     onTextAreaFocused(e: any) {
         mobx.action(() => {
             this.isFocused.set(true);
```
esimkowitz, line 113: Do we need a local isFocused state? We should just be able to use the auxViewFocus global var to track this for all the views

src/app/workspace/cmdinput/aichat.tsx
```diff
@@ -243,7 +246,8 @@ class AIChat extends React.Component<{}, {}> {
                     this.isFocused.get() ||
                     GlobalModel.inputModel.hasFocus() ||
                     (GlobalModel.getActiveScreen().getFocusType() == "input" &&
-                        GlobalModel.activeMainView.get() == "session")
+                        GlobalModel.activeMainView.get() == "session" &&
+                        GlobalModel.inputModel.getActiveAuxView() == appconst.InputAuxView_AIChat)
```
esimkowitz, line 250: Maybe we should make this logic into a function in inputModel, (`isAuxViewFocused(auxView: InputAuxViewType)`). This could use the logic above and return true or false depending on if the focus is applied.

src/app/workspace/cmdinput/historyinfo.tsx
```diff
@@ -174,7 +174,7 @@ class HistoryInfo extends React.Component<{}, {}> {
     handleItemClick(hitem: HistoryItem) {
         const inputModel = GlobalModel.inputModel;
         const selItem = inputModel.getHistorySelectedItem();
-        inputModel.setAuxViewFocus(false);
+        inputModel.setAuxViewFocus(!inputModel.getAuxViewFocus());
```
esimkowitz, line 177: I think this could cause issues, why is setting it to false not working?

src/app/workspace/cmdinput/textareainput.tsx
```diff
@@ -604,17 +604,53 @@ class TextAreaInput extends React.Component<{ screen: Screen; onHeightChange: ()
                 }
             }
         }
-        const isHistoryFocused = auxViewFocused && inputModel.getActiveAuxView() == appconst.InputAuxView_History;
+
+        const renderCmdInputKeybindings = mobx
+            .computed(() => {
+                const inputModel = GlobalModel.inputModel;
```
esimkowitz, line 610: Same as above, we should make this a reusable function that can accept a aux view type and return whether the keybindings should be set.

src/models/input.ts
```diff
@@ -691,6 +691,7 @@ class InputModel {
 
     openAIAssistantChat(): void {
         this.setActiveAuxView(appconst.InputAuxView_AIChat);
+        this.setAuxViewFocus(true);
```
esimkowitz, line 694: setActiveAuxView already sets auxViewFocus

src/models/model.ts
```diff
@@ -752,10 +752,12 @@ class Model {
                 setTimeout(() => {
                     // allows for the session view to load
                     this.inputModel.giveFocus();
+                    this.inputModel.setAuxViewFocus(false);
```
esimkowitz, line 755: replace the call to giveFocus with setAuxViewFocus. it will call giveFocus

src/models/model.ts
```diff
@@ -752,10 +752,12 @@ class Model {
...
                 }, 100);
             })();
         } else {
             this.inputModel.giveFocus();
+            this.inputModel.setAuxViewFocus(false);
```
esimkowitz, line 760: replace the call to giveFocus with setAuxViewFocus. it will call giveFocus

src/app/workspace/cmdinput/textareainput.tsx
MrStashley: great idea, addressed

src/app/workspace/cmdinput/aichat.tsx
MrStashley: I don't think we do, addressed

src/models/input.ts
```diff
@@ -500,6 +500,29 @@ class InputModel {
         this.giveFocus();
     }
 
+    shouldRenderAuxViewKeybindings(view: InputAuxViewType): boolean {
```
MrStashley, line 503: Here's the function that I implemented that contains logic for deciding when we should render keybindings

esimkowitz:
Sorry for the delay, this looks good to me!!

PR #578, opened 2024-04-17
Block store

MrStashley:
This contains my implementation of the block store, as described in https://www.notion.so/wavetermdev/Block-Store-30465cad3a7c4f018da5143c4591663f

In this pr is a simple way to send and receive data through a database

I have implemented the base functionality as well as quite a few tests to make sure that everything works 
 
There are a few methods that have yet to be implemented, but theoretically they should be implemented as calls to the other functions, ie append should just be a call to WriteAt

I think I will make it so that a writeAt call sets a go routine that initiates the flush 

We need to be careful with reads and writes. If we're caching continuous writes every second or so, it would be very easy to clobber any data that was sent over by the front end. I feel like in practice if we're going to be doing both reads and writes, we should have 2 files, one for read and one for write. either that, or have a tcp like protocol where we send acks

This doesn't affect anything so it can safely be merged whenever. I don't want this pr to stagnate like file view, so I'm happy to write multiple prs for this

I know it's a bit verbose, down to hop on a call to help with the review process


wavesrv/pkg/blockstore/blockstore.go
```diff
@@ -0,0 +1,389 @@
...
+	CollapseIJson(ctx context.Context, blockId string, name string) error
+	WriteMeta(ctx context.Context, blockId string, name string, meta FileMeta) error
+	DeleteFile(ctx context.Context, blockId string, name string) error
+	DeleteBlock(ctx context.Context, blockId string) error
+	ListFiles(ctx context.Context, blockId string) []FileInfo
+	FlushCache(ctx context.Context) error
+	GetAllBlockIds(ctx context.Context) []string
+}
+
+var cache map[string]*CacheEntry = make(map[string]*CacheEntry)
```
MrStashley, line 68: currently I use global memory like in sstore.go, but since BlockStore is an interface I could make a `BlockStoreType` with all of the data. Then we could have multiple block stores, like if we wanted a different one for web communication or something like that that might be a use case. Should I do that?

wavesrv/pkg/blockstore/blockstore.go
```diff
@@ -0,0 +1,389 @@
...
+		tx.Exec(query, fileInfo.BlockId, fileInfo.Name, fileInfo.Opts.MaxSize, fileInfo.Opts.Circular, fileInfo.Size, fileInfo.CreatedTs, fileInfo.ModTs, metaJson)
+		return nil
+	})
+	if txErr != nil {
+		return fmt.Errorf("Error writing file %s to db: %v", fileInfo.Name, txErr)
+	}
+	return nil
+}
+
+func WriteFileToDB(ctx context.Context, fileInfo FileInfo) error {
```
MrStashley, line 86: Should prob move these to the dbops file

MrStashley: also can combine this function and the above using REPLACE syntax

wavesrv/pkg/blockstore/blockstore.go
```diff
@@ -0,0 +1,449 @@
...
+		return 0, 0, fmt.Errorf("Cache entry not in cache")
+	}
+	cacheEntry.Lock.Lock()
+	defer cacheEntry.Lock.Unlock()
+	var bytesWritten = 0
+	blockLen := len(block.data)
+	fileMaxSize := cacheEntry.Info.Opts.MaxSize
+	maxWriteSize := fileMaxSize - (int64(cacheNum) * MaxBlockSize)
+	numLeftPad := int64(0)
+	if pos > blockLen {
```
MrStashley, line 141: So in the case where the offset is greater than the file size, like if we call WriteAt(100) on a new file, the block store will left pad 0's. This creates some options and some potential concerns. 

First of all, do you think I should left pad 0's, or should I just return an error if the write at offset is greater than the file size? 

Currently, what I do is I calculate the left pad based on when I am writing to a cache block. So I only left pad in the current cache rather than the whole file. I add the amount of left pad bytes to the bytesWritten return value. This could lead to some ambiguity in the bytesWritten return value

I have one test where I create a circular file whose max size is 2 blocks
I write an 11 byte message to an offset 4 bytes from the end of the file. 
the code first writes 4 bytes at the end of the second cache, and it left pads enough 0s to fill the entire cache block except for the last 4 bytes. 
Then, the code writes the last 7 bytes of the message to the front of the first cache block. There is no need to left pad any bytes to the cache, so the first cache block ends up being only 7 bytes long, while the second cache block is BlockSize bytes long. The bytesWritten value returned by WriteAt is MaxBlockSize + 7, although the size of the file is now 2 * MaxBlockSize, but the actual stored size of the first cache is only 7 bytes

When the user calls read with offset 0, reading from the beginning of the file, the returned bytesRead value will be only 7 bytes, but it would normally read 0s until the end of the buffer if the code had filled the first cache block with 0s 

The other thing that I could do is left pad the entire file with 0s up to the write position, rather than just the cache, so now there is no potential for ambiguous behavior, but do we think that would be a significant performance hit? 

Say I want to write at offset 5gb in an empty file, I could be writing 5gb of left pad 0s, where if I only left pad the current cache the max left padding that I would be doing is BlockSize

Also, as a secondary general question, should I return the amount of left pad bytes in the return value of WriteAt? I'm thinking maybe it makes sense to just return the actual bytes written so that there is no ambiguity in return value 

Overall, my thoughts on this are that I think we should return an error if offset is greater than file size unless we have a good use case for it, and if so, I think we should keep the cache optimization and also not return the left pad bytes in the return value of WriteAt

wavesrv/pkg/blockstore/blockstore_test.go
```diff
@@ -0,0 +1,913 @@
...
+	bytesRead, err := ReadAt(ctx, "test-block-id-sync", "file-1", &readBuf, 0)
+	if err != nil {
+		t.Errorf("Read Error: %v", err)
+	}
+	log.Printf("read buf : %v", readBuf)
+	SimpleAssert(t, bytesRead == numWorkers, "Correct num bytes read")
+	CleanupName(t, ctx, "test-block-id-sync")
+}
+
+func TestWriteAtSyncMultiple(t *testing.T) {
```
MrStashley, line 810: Sync test for WriteAt
Hopefully this should be good enough to be fairly sure that our sync assumptions are correct

wavesrv/pkg/blockstore/blockstore.go
```diff
@@ -0,0 +1,449 @@
...
+				}
+			} else {
+				return bytesRead, fmt.Errorf("Read from cache error: %v", err)
+			}
+		}
+	}
+	return bytesRead, nil
+}
+
+func AppendData(ctx context.Context, blockId string, name string, p []byte) (int, error) {
```
MrStashley, line 438: I lock the entirety of AppendData, because append needs to get the offset for the end of the file and then update the end of the file without the end of the file changing

wavesrv/pkg/blockstore/blockstore.go
MrStashley: Update on this - I went with the cache block only left pad approach, and I made it so left pad bytes are not returned in the return value of the WriteAt call

There is ambiguity in the ReadAt return value still. If you read from a cache block that hasn't been left padded yet, but is not the last block in the file, and you try to read some bytes that "should" be 0 but are currently empty, read at doesn't return those bytes 

Psuedocode Example - lets say MaxBlockSize is 5
WriteAt(dataToWrite: "test", offset: 6)
After the write, the cache looks like: 
Cache Entry File Size: 10 
Block1 { data: [], size: 0}
Block2 {data: [0 t e s t, size: 5}

readBuf = make([]byte, 5) 
bytesRead = ReadAt(buf: readBuf, offset: 0)
bytesRead is 0 
readBuf is unchanged
really readBuf should be filled with 0s and bytesRead should be 5 

Should I make a special case for this?

MrStashley: Currently I have no special case for this 

this is probably the only "open" for me, unless any reviewers have any issues 

everything else I would say I'm pretty comfortable with

PR #579, opened 2024-04-17
fix: inline text edit button use onMouseDown

LawlietLi:
This PR fix a issue that after edit text in `InlineSettingsTextEdit` click `Confirm` button is not work.

截屏2024-04-17 22 52 03

Because of `onBlur` has high priority than `onClick` when click `Confirm` button always execute `handleBlur` first, so edit will be canceled. `onMouseDown` has high priority than `onBlur`, can fix this issue.


sawka:
Thanks for submitting!  Will review this to be merged for the upcoming patch release (v0.7.3).

PR #590, opened 2024-04-23
Open AI Timeout setting and alias for all settings

MrStashley:
Added a timeout setting for our ai and added aliases for all of the ai settings. 

Note that the timeout that can be set is the packet timeout, ie the timeout when waiting for a server response. The entire stream has a timeout of 5 minutes, I'm thinking we should up that timeout as well, perhaps to 10 minutes 

Or should I have a setting for that timeout as well? I think the packet response one is the one that we really should be setting, but not 100% sure

### DIFF
--- a/wavesrv/pkg/cmdrunner/cmdrunner.go
+++ b/wavesrv/pkg/cmdrunner/cmdrunner.go
@@ -2692,8 +2692,6 @@ func getCmdInfoEngineeredPrompt(userQuery string, curLineStr string, shellType s
 }
 
 func doOpenAICmdInfoCompletion(cmd *sstore.CmdType, clientId string, opts *sstore.OpenAIOptsType, prompt []packet.OpenAIPromptMessageType, curLineStr string) {
-	var hadError bool
-	log.Println("had error: ", hadError)
 	ctx, cancelFn := context.WithTimeout(context.Background(), OpenAIStreamTimeout)
 	defer cancelFn()
 	defer func() {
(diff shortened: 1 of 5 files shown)


wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -5784,6 +5787,16 @@ func validateFontFamily(fontFamily string) error {
 	return nil
 }
 
+func CheckOptionAlias(kwargs map[string]string, alias1 string, alias2 string) (string, bool) {
```
sawka, line 5790: may as well make this a varadic function.  change to `aliases ...string`

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -5873,7 +5886,7 @@ func ClientSetCommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (sc
 			return nil, fmt.Errorf("error updating client openai api token: %v", err)
```
sawka, line 5886: please change these messages as well.  should just be "ai api token" instead of "openai"

wavesrv/pkg/cmdrunner/cmdrunner.go
```diff
@@ -5999,6 +6029,7 @@ func ClientShowCommand(ctx context.Context, pk *scpacket.FeCommandPacketType) (s
 	buf.WriteString(fmt.Sprintf("  %-15s %d\n", "openaimaxtokens", clientData.OpenAIOpts.MaxTokens))
```
sawka, line 6029: change these as well to remove "openai" and just say "ai"

gonzolively:
Not sure if i'm missing something or not, but when I pulled this branch and built wave dev I do not see any renamed attributes in `client:show`.

Also, looks like there is an error with the new timeout option.

gonzolively:
Not sure if i'm missing something or not, but when I pulled this branch and built wave dev I do not see any renamed attributes in `client:show`.

Also, looks like there is an error with the new timeout option as the value is showing `%!s(int=0)`.

Screenshot 2024-04-23 at 15 56 07

PR #594, opened 2024-04-23
Small aux view fixes that make cmdinput focus work a bit better

MrStashley:


src/models/input.ts
```diff
@@ -691,6 +688,7 @@ class InputModel {
 
     openAIAssistantChat(): void {
         this.setActiveAuxView(appconst.InputAuxView_AIChat);
+        this.setAuxViewFocus(true);
```
esimkowitz, line 691: setActiveAuxView already calls setAuxViewFocus(true)

esimkowitz: makes sense!

PR #612, opened 2024-04-26
Attempt 2 to resolve security warning

esimkowitz:
This is the second attempt to resolve the "Uncontrolled data in path expression" warning. I am also using this opportunity to replace `path` with `filepath` in the `main-server.go` file.


MrStashley:
Looks good to me!

PR #615, opened 2024-04-27
Fixes a regression that would unfocus from the history input

esimkowitz:
This fixes a regression where the history auxiliary cmd view would come into focus, give focus to the history input field, and then immediately it would get reset to the cmdinput field. This was due to an erroneous binding of the cmdInput.onSelect event, which was set to give focus back to the textarea. 

This also fixes a regression with the new OverlayScrollbarsComponent, which was overriding the scroll position of the history auxview once it finished initializing. I have updated the aux view to allow binding of a handler to the initialization event. This will ensure the scroll position is only applied once the initialization is complete.

This also updates the history auxview to utilize the `flex-direction: column-reverse`, which lets us avoid reversing the whole array in JS, which can be expensive in screens with a lot of history.

### DIFF
--- a/src/app/workspace/cmdinput/historyinfo.tsx
+++ b/src/app/workspace/cmdinput/historyinfo.tsx
@@ -154,17 +154,18 @@ class HistoryInfo extends React.Component<{}, {}> {
     lastClickTs: number = 0;
     containingText: mobx.IObservableValue<string> = mobx.observable.box("");
 
-    componentDidMount() {
+    /**
+     * Handles the OverlayScrollbars initialization event to set the scroll position without it being overridden.
+     */
+    @boundMethod
+    handleScrollbarInitialized() {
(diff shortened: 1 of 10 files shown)


src/app/workspace/cmdinput/historyinfo.tsx
```diff
@@ -161,7 +161,10 @@ class HistoryInfo extends React.Component<{}, {}> {
             hitem = inputModel.getFirstHistoryItem();
         }
         if (hitem != null) {
-            inputModel.scrollHistoryItemIntoView(hitem.historynum);
+            // This is a hack to ensure that any other scrolling instru
```
esimkowitz, line 164:
```suggestion
            // This is a hack to ensure that our call to scroll to the current item is not overwritten. 
```

PR #633, opened 2024-05-02
aichat sidebar

adred:


src/app/sidebar/aichat.tsx
```diff
@@ -0,0 +1,322 @@
...
+        }).catch((_) => {});
+    }
+
+    getLinePos(elem: any): { numLines: number; linePos: number } {
+        const numLines = elem.value.split("\n").length;
+        const linePos = elem.value.substr(0, elem.selectionStart).split("\n").length;
+        return { numLines, linePos };
+    }
+
+    @mobx.action.bound
```
esimkowitz, line 196: you'll need to add a constructor that calls `mobx.makeObservable(this)` in order for these attributes to work

src/app/sidebar/aichat.tsx
```diff
@@ -0,0 +1,322 @@
...
+                        {this.renderChatMessage(chitem)}
+                    </For>
+                </div>
+            </OverlayScrollbarsComponent>
+        );
+    }
+}
+
+@mobxReact.observer
+class AIChat extends React.Component<{}, {}> {
```
esimkowitz, line 159: Can we not reuse more from the aichat auxview or maybe make that a component somewhere else that both views can utilize? Same for the keybindings

sawka:
red, you're going to have to resolve the conflicts here -- including migrate to use clsx (instead of classnames) and the new "yarn modern".

sawka:
closing, since I think this is the old PR.  I don't want to get confused.

PR #639, opened 2024-05-02
ai chat selection fix

MrStashley:


esimkowitz:
I would use PointerDown/PointerUp instead if you can, this will let it work for touch events too

sawka:
I think you should look at whether there is a selection (the user has highlighted some text).  so you can still use onClick, but check:
```
const sel = window.getSelection();
if (isBlank(sel?.toString())) {
}
```

PR #649, opened 2024-05-04
enable self-signed TLS between client/server

sawka:
create an in-memory self-signed tls cert on the golang side and trust it on the electron side.  enables http/2 protocol for increased throughput (simultaneous connections)


sawka:
need to suppress error message in electron log:
```
[89461:0503/215419.465397:ERROR:ssl_client_socket_impl.cc(879)] handshake failed; returned -1, SSL error code 1, net_error -202
```

also potential issue with startup time.  might have to switch to a static cert instead of generating it on the fly.

esimkowitz:
Closed as no longer relevant

PR #731, opened 2024-07-26
Use gpt-4o-mini by default

kamushadenes:
OpenAI [recently announced](https://openai.com/index/gpt-4o-mini-advancing-cost-efficient-intelligence/) the launch of GPT 4o mini, a small LLM model that's both smarter and cheaper than GPT-3.5 Turbo, meaning there's no reason not to use it by default.

This simple PR replaces all occurences of gpt-3.5-turbo with gpt-4o-mini.


esimkowitz:
This looks good to go, just waiting because we are making an update to our backend to set gpt4o-mini as the default there too!

esimkowitz:
Closing as no longer relevant wIth the new codebase

PR #732, opened 2024-07-26
Listen on random ports

### LINKED ISSUE #369: Concurrent multi-user not supported
It doesn't seem possible to run waveterm concurrently with different users.

1. Start waveterm with one user.
2. Change to a different OS user
3. Start waveterm

The app starts as a black window. Looking in the dev console,  this error message is repeated multiple times

```
WaveTerm v0.6.3 20240220-021439
get-client-data:1 
        
        
       Failed to load resource: the server responded with a status of 500 (Internal Server Error)
model.ts:4614 [error] calling get-client-data Error: Bad status code response from fetch 'http://127.0.0.1:1619/api/get-client-data': code=500 Internal Server Error
x-authkey header is invalid
    at util.ts:39:15

```

We often have dev processes running with one user and switch to a different user for testing reasons.  It's possible to alter the test workflow to sandbox things differently using remote connects but this prevents the tool from being a drop-in replacement for the standard terminal.

**Desktop (please complete the following information):**
 - OS: MacOS
 - Version 14.2.1

kamushadenes:
Fixes #369


esimkowitz:
This looks awesome! We should use this for both dev and prod endpoints though, no reason to keep them separate anymore if it's all random!

kamushadenes:
Figured I would remove dev endpoints all together.

esimkowitz:
Closing as no longer relevant with the new codebase

PR #734, opened 2024-07-26
Improve code quality and legibility

kamushadenes:
This PR intends to improve Golang code quality and legibility to facilitate future changes.

A non-exhaustive list of changes is:

- Remove a ton of dead code / unused functions and vars in order to simplify review and auditing
- Simplify some nested ifs that made the code hard to read (there are many left though)
- Where applicable, replace `if-else` with `switch-case`
- Replace many instances of `strings.Index()` checks with `strings.Contains()`
- Replace `strings.HasSuffix() + Trim` with non-conditional `strings.TrimSuffix()`
- Use `var <name> <type>` instead of `<name> := <type>(zero value)`
- Replace var names that conflicted with reserved keywords (len, close, etc...)
- Properly handle ineffectual assignments (like errs that weren't checked etc)
- Remove unnecessary `fmt.Sprintf()` on non-formatting strings
- Remove a bunch of redundant returns
- Remove a bunch of unnecessary / redundant conversions
- Simplify expressions such as `x = x + y` => `x += y`
- Replace `string(r.Bytes())` with `r.String()` in *bytes.Buffer instances.
- Format files using `go fmt`
- Replace some strings with consts, such as `POST` => `http.MethodPost` and `200` => `http.StatusOK` (there are still many repeated strings that could become consts)
- Use `append(slice, newSlice...)` instead of a for loop
- Use raw strings in `regex.MustCompile`
- Fix instances of `== true` and `if cond == true ... return true ... else return false`
- And others

There are still many improvements to be made, but this PR was already too big so I'll make separate PRs after this one (hopefully) gets merged.

@esimkowitz I need help testing this. It *shouldn't* change anything (besides perhaps throwing some errors that were previously incorrectly ignored), but better safe than sorry.

### DIFF
--- a/waveshell/pkg/base/base.go
+++ b/waveshell/pkg/base/base.go
@@ -51,9 +51,9 @@ var ensureDirCache = make(map[string]bool)
 var baseLock = &sync.Mutex{}
 var DebugLogEnabled = false
 var DebugLogger *log.Logger
-var BuildTime string = "0"
+var BuildTime = "0"
 
-var ProcessType string = ProcessType_Unknown
+var ProcessType = ProcessType_Unknown
 
 type CommandFileNames struct {
 	PtyOutFile    string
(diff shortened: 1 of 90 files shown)


esimkowitz:
Closing as we've made significant changes to Wave with v0.8 that make these changes no longer relevant

PR #752, opened 2024-08-06
Add Linux arm64 build

### LINKED ISSUE #745: Provide linux arm64 build?
**Is your feature request related to a problem? Please describe.**
while it's not as popular as amd64 there are many machines that run arm64 linux
for example raspberry pi, m1 m2 ... apple devices, arm64 chromebooks and likely some more stuff

**Describe the solution you'd like**
it would be nice to include a build for these platforms too?
if it won't cost too much effort i see no reason not to do that

kamushadenes:
Closes #745

### DIFF
--- a/.github/workflows/build-helper.yml
+++ b/.github/workflows/build-helper.yml
@@ -16,6 +16,10 @@ jobs:
                       arch: "amd64"
                       runner: "ubuntu-latest"
                       scripthaus: "build-package-linux"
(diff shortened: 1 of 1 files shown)


esimkowitz:
Updating to use our Linux ARM runner

esimkowitz:
Thanks @kamushadenes for this! I will make sure we have an ARM build in our next release!

PR #767, opened 2024-08-25
Containerize project : configure dockerfile and compose files

### LINKED ISSUE #740: Containerize this Project for Easier Contribution
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when trying to set up the development environment for this project. The process can be time-consuming and error-prone, particularly when dealing with different dependencies and system configurations.

**Describe the solution you'd like**
I propose containerizing this project using Docker. By providing a pre-configured Docker container, developers can quickly and easily set up a consistent development environment. This would streamline the onboarding process for new contributors and ensure that everyone is working with the same setup, reducing the likelihood of environment-related issues.

sujal-goswami:
resolve #740 

For user to build this project using docker they need to fulfil the prerequisites:

## Prerequistes
-  Docker Desktop or docker should be installed
-  WSL - Windows Subsystem for Linux  (only for windows users)
- should have `X server` installed in host machine
if not have then can download from command `sudo apt-get install x11-xserver-utils`.
to check its correct installation run `xhost` this should give output as 
```
access control disabled, clients can connect from any host
LOCAL:
SI:localuser:wslg
```
## demo image
wave

## References
[https://github.com/cypress-io/cypress-documentation/issues/5816](https://github.com/cypress-io/cypress-documentation/issues/5816)
[--no-sandbox error](https://ask.replit.com/t/get-error-by-command-chromium-browser-no-sandbox/14514/2)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+FROM ubuntu:latest
+WORKDIR /app
+
+RUN apt-get update -y && \
+    apt-get upgrade -y && \
+    apt-get install -y golang-go nodejs npm && \
+    npm install -g corepack && \
+    corepack enable && \
+    yarn install && \
+    apt-get install -y libgtk2.0-0t64 libgtk-3-0t64 libgbm-dev libnotify-dev libnss3 libxss1 libasound2t64 libxtst6 xauth xvfb && \
+    export DISPLAY=:0 && \
+    service dbus start
+
+RUN git clone https://github.com/scripthaus-dev/scripthaus.git
+WORKDIR /app/scripthaus
+
+RUN CGO_ENABLED=1 go build -o scripthaus cmd/main.go && \
+    cp scripthaus /usr/local/bin
+
+WORKDIR /app
+RUN mkdir waveterm
+
+WORKDIR /app/waveterm
+COPY . .
+RUN yarn cache clean
+RUN yarn
+RUN scripthaus run electron-rebuild
+RUN scripthaus run build-backend
+RUN scripthaus run webpack-watch
+
+RUN useradd -m -s /bin/bash wave
+USER wave
+
+CMD [ "scripthaus", "run", "electron" ]
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,5 +31,3 @@ RUN scripthaus run webpack-watch
 RUN useradd -m -s /bin/bash wave
 USER wave
 
-CMD [ "scripthaus", "run", "electron" ]
-
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,19 @@
+{
+    "name": "Wave Terminal Dev Container",
+    "dockerComposeFile": "../docker-compose.yml",
+    "service": "wave",
+    "workspaceFolder": "/app/waveterm",
+    "customizations": {
+        "vscode": {
+            "settings": {
+                "terminal.integrated.shell.linux": "/bin/bash",
+                "extensions": [
+                    "golang.go",
+                    "dbaeumer.vscode-eslint",
+                    "esbenp.prettier-vscode"
+                ]
+            }
+        }
+    },
+    "remoteUser": "wave"
+}
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,9 +13,6 @@ logs
 
 
 
-# Ignore Docker files
-Dockerfile
-.dockerignore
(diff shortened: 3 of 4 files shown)


esimkowitz:
Thank you for setting this up! Since this is for a dev container, I think the .dockerignore doesn't need to be as restrictive, since the developer may care about the .md, config, and Yarn files. Otherwise it looks good to me!

esimkowitz (changes requested):
Based on your issue, I assumed you meant to create a docker container for development purposes, usually for something like this you would want to add the VSCode tooling for Dev Containers and leave the repo data inside the final container. (see [Developing inside a Container](https://code.visualstudio.com/docs/devcontainers/containers)

I'm not sure I see a point to just running our app in a container, since it won't have access to the rest of your system, which kind of defeats the purpose of a terminal app.

esimkowitz:
Could you verify that this complies with the guidelines for the dev containers in the guide I linked ([Developing inside a Container](https://code.visualstudio.com/docs/devcontainers/containers))? This one may also be helpful: [Create a development container](https://code.visualstudio.com/docs/devcontainers/create-dev-container#_path-to-creating-a-dev-container). I think you can skip setting a CMD to start on container run, since this is not necessary for developing inside a container. Also I realize the suggested changes I made left a bunch of new lines, could you clean those up?

sujal-goswami:
Sure !

sujal-goswami:
now I think it is good to go; now, the user can use the VS Code dev container feature for seamless development in the container with pre-install go , eslint and prettier extensions in it.

esimkowitz:
I don't think you need to build the app at all in the dockerfile, since it's fast enough. Also, because webpack-watch is blocking, the dockerfile never completes. I moved this to the "postStartCommand" in devcontainer.json, but I'm still unable to launch the app, as it's complaining about DBus not being started. I tried moving the `service dbus start` command to the postStartCommand too, but it's still not working. What host OS are you using to test this? I'm on macOS 14.

I'm concerned that if the devcontainer isn't reliable on all platforms or requires additional hoops to get working, it won't be that much better than just installing the deps directly. Aside from scripthaus, we don't have any crazy dependencies. We're also in the process of a pretty large rearchitecting and we're removing the scripthaus dependency as part of this, so our dependency story will get a lot simpler.

sujal-goswami:
I don't think the host OS is the issue here. It's more likely an issue with the dbus service. I encountered a similar dbus error during my testing but resolved it by running `service dbus start` on the machine. You can refer to this [link](https://ask.replit.com/t/get-error-by-command-chromium-browser-no-sandbox/14514/2) for more details. By the way, I'm using Windows as the host OS, so it's odd if the OS is indeed the problem. The rearchitecting plan sounds great! Simplifying the dependencies will definitely streamline the process.

klmcwhirter:
FWIW, I did something similar as this PR - but, my goal was to build waveterm in a container - not run it in the container.

I purposely do not install multiple tool chains, project specific dependencies, etc. on localhost.

FYI, I am running on Fedora 41 and typically use VS Code (although I have been using `emacs -nw` again lately.

I concur that .devcontainer might be of dubious value, and there are several missing extensions that I typically use for web dev (vscode-tailwindcss, etc.). So I would probably just create my own.

In my case, I chose fedora:41 as the base OCI image because I wanted to be sure the resulting binary would run on my system.

That is not an issue when running in a container. If I was looking for containerization, I would prefer flatpak as an execution env because it is a more natural fit.

I suspect there may be a number of personal preferences out there.

But agree with @esimkowitz - thanks for putting in the effort. I am sure there are devs out there who can benefit.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    apt-get upgrade -y && \
+    apt-get install -y golang-go nodejs npm && \
+    npm install -g corepack && \
+    corepack enable && \
+    yarn install && \
+    apt-get install -y libgtk2.0-0t64 libgtk-3-0t64 libgbm-dev libnotify-dev libnss3 libxss1 libasound2t64 libxtst6 xauth xvfb && \
+    export DISPLAY=:0 && \
+    service dbus start
+
+RUN git clone https://github.com/scripthaus-dev/scripthaus.git
```
klmcwhirter, line 14: Someone probably already has commented on this ...

But is this the github project you really intended to merge? note scripthaus-dev ...

esimkowitz:
Closing this PR for now, sorry we couldn't get it merged

Issue #707, opened 2024-07-01
Unable to connect to SSH host


ekool:
I have a host I am unable to add. No matter what type of auth or settings I choose, I get this error:

wave> error, ssh cannot connect to client: ssh: handshake failed: remote host identification has changed

Connecting to this host via SSH via regular terminal works fine. No other error is given. (latest Mac client installed via homebrew).

jimmyshe:
same situation here

zerorendan:
same here

FrankSchoene:
the same

joergmschulz:
Please use the thumbs-up icon for the first post if you can't contribute anything but a me-too. Thanks! This might bost the probability we get an answer

oneirocosm:
Unfortunate to see this is still happening. Since we are using the golang ssh/knownhosts library, there have been some cases where our parsing of knownhosts is incorrect.  Unfortunately, I haven't been able to reproduce it myself, but it's widespread enough that there must be a straightforward way to reproduce it.  To help narrow this down:
- Do any of you have particular keywords in your ~/.ssh/config file that are related to parsing known hosts?
- Also, what does your ~/.ssh/known_hosts look like? Are there multiple publickeys per host listed? Are the keys being hashed in a specific way?

In an absolute worst case scenario, you could temporarily get around the issue by deleting the host from your known_hosts file and add it again while connecting with wave. I don't think this is a satisfactory solution, so I'll be looking for a better system on our end.

zerorendan:
> Unfortunate to see this is still happening. Since we are using the golang ssh/knownhosts library, there have been some cases where our parsing of knownhosts is incorrect. Unfortunately, I haven't been able to reproduce it myself, but it's widespread enough that there must be a straightforward way to reproduce it. To help narrow this down:
> 
> * Do any of you have particular keywords in your ~/.ssh/config file that are related to parsing known hosts?
> * Also, what does your ~/.ssh/known_hosts look like? Are there multiple publickeys per host listed? Are the keys being hashed in a specific way?
> 
> In an absolute worst case scenario, you could temporarily get around the issue by deleting the host from your known_hosts file and add it again while connecting with wave. I don't think this is a satisfactory solution, so I'll be looking for a better system on our end.
![Screenshot 2024-07-05 at 11 34 29 AM](https://github.com/wavetermdev/waveterm/assets/8052614/169f88e8-6b9e-4606-99a4-1179e79d69fd)
[...]

PR #770, opened 2024-08-28
Integrate Skeema Knownhosts Fix

oneirocosm:
This fix makes it possible to differentiate between keys when multiple are provided by the remote server. It does not solve the case of multiple keys of the same type being shared, but it handles multiple keys of different types being shared, which is much more common. This should address most issues similar to #707 .


oneirocosm:
merge issue. opening a new pr since it's simpler

PR #860, opened 2024-09-26
Plot Variety

oneirocosm:
Adds the ability to switch between several different plot types and overlay them on top of one another. This adds
- original cpu plot is still present
- plots for individual cpus
- total memory
- memory used
- memory available
- memory free


sawka:
we have a different branch that is adding this functionality now

PR #912, opened 2024-10-01
fix args issue in menu stories

adred:


adred:
already in another branch

Issue #925, opened 2024-10-02
`wsh edit file` needs to create a new file if file does not exist


ToontjeM:
**Is your feature request related to a problem? Please describe.**
Currently `wsh edit` can only edit existing files. So creating a new file is a two-step process of `touch file` and `wsh edit file`.

**Describe the solution you'd like**
`wsh edit file` should create file if file does not exist.

esimkowitz:
Fixed with #963

PR #963, opened 2024-10-05
Open Text File if File Does Not Exist

oneirocosm:
This will open new files as empty text files for editing.


sawka:
tagging issue #925

PR #1101, opened 2024-10-22
sub-blocks and parentoref

sawka:


sawka:
canceling this PR -- integrated directly into vdom4

PR #1106, opened 2024-10-23
Various toolkit elements

adred:


adred:
already in https://github.com/wavetermdev/waveterm/pull/1148

PR #1159, opened 2024-10-28
feat: add checkmark next to current connection

oneirocosm:
The is specifically for the typeahead dropdown for connections.


sawka:
regression -- cannot move up/down arrow to the last connection when there is a search

oneirocosm:
should be fixed with the latest commit

PR #1180, opened 2024-10-31
vdom 7

sawka:


sawka:
* component functions can now take structs as props (will remarshall correctly)
* implemented file transfer vdom:// protocol -- supports ``
* transform style attributes
* implement svg tags + id attributes
* other bug fixes and layout issues

Issue #1167, opened 2024-10-30
[Bug]: v0.9.0 - Tabs getting deleted after auto update


C0sm0cats:
### Current Behavior

I've noticed that every time I open and close the application, the tab in my browser is renamed by incrementing T number . Additionally, if I open multiple tabs, upon reopening the application, it only restores one tab and increments the number on the title. the application does not save its state, resulting in a loss of unsaved progress.
![image](https://github.com/user-attachments/assets/24b9a3a9-c869-4f85-995f-6feca89eb28f)

### Expected Behavior

The tab title should remain consistent each time the application is opened and closed, without any incremental numbering.

### Steps To Reproduce

Open the application.
Open a new tab or several tabs
Close the application.
Reopen the application.
Observe the tab title.

### Wave Version

v0.9.0

### Platform

Linux

### OS Version/Distribution

manjaro

### Architecture

x64

### Questionnaire

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

Exc4l1buR:
I'm experiencing the same thing Tuxedo OS (based on Ubuntu).
Though it does seem to keep the state if you crash it by trying to open the view menu from the main menu.

crush2000:
I have the same issue on ZorinOS 17.2 (based on Ubuntu)

C0sm0cats:
yes and i guess you have this bug also => https://github.com/wavetermdev/waveterm/issues/1152

Exc4l1buR:
yup, also i dont know if this is relevent but im using wayland on my system

crush2000:
Rollback to 0.8.12 didn't work for me. First run on default set of widgets is ok but form second and next it's broken like in 0.9.0. Snap also is broken.

esimkowitz:
We believe that this was a race condition between our updater and our window manager. During normal use, if you close a window we delete it from the backend, including all the tabs (with the exception of the last tab). This should not happen during an update, though, as we set a flag before shutdown to bypass this step. We believe that with our refactor of our window manager for 0.9, the timing for the event that triggers the window close changed and the event was firing before the status flag was set by the updater. This should now be fixed as we added a delay after setting this flag before we actually trigger the install.

Going forward, this will be fixed more robustly by us reintroducing workspace support and adding automatic archiving of old windows. We were hopeful workspaces would make it into this last release, but it needs some more polishing.

[...]

PR #1181, opened 2024-11-01
make sure the user affirmatively acknowledged window deletion before deleting, fix for #1167

sawka:


sawka:
make sure the user affirmatively acknowledged window deletion before deleting, fix for #1167

PR #1206, opened 2024-11-06
VDom 10

sawka:
get RefOperations and RefUpdates working.  also implement a <canvas> API that can be called using RefOperations.


sawka:
* implemented canvas operations with refs
* fix rehype/markdown memory leak for demo

PR #1278, opened 2024-11-14
use color var for connection status overlay

adred:


frontend/app/block/block.less
```diff
@@ -415,7 +415,7 @@
             &.block-no-highlight,
             &.block-preview {
                 .block-mask {
-                    border: 2px solid rgba(255, 255, 255, 0.1) !important;
+                    border: 2px solid rgb(from var(--border-color) r g b / 1%) !important;
```
esimkowitz, line 418: should be 10% i think

adred: addressed

esimkowitz (approved):

PR #1325, opened 2024-11-20
rating notification

adred:


frontend/app/notification/ratingbubble.less
```diff
@@ -0,0 +1,36 @@
...
+
+.notification-rating-bubble {
+    position: relative;
+    display: flex;
+    width: 432px;
+    padding: 16px 24px 16px 16px;
+    align-items: flex-start;
+    gap: 12px;
+    border-radius: 8px;
+    border: 0.5px solid rgba(255, 255, 255, 0.12);
```
esimkowitz, line 12: Can you use the colors from the themes variables in this file?

esimkowitz: I think you'll also need to manually update this to scss

adred: addressed

sawka:
too old

PR #1346, opened 2024-11-22
stronger callouts for the 'clear' key

sawka:


esimkowitz:
maybe move the info to the top

PR #1383, opened 2024-12-04
New Connections Configs

oneirocosm:
This adds the following connections changes:
- connections can be hidden from the dropdown in our internal connections.json config
- `wsh ssh` -i will write identity files to the internal connections.json config for that connection
- the internal connections.json config will also be used to get identity files when connecting
- the internal connections.json config allows setting theme, fontsize, and font for specific connections
- successful connections (including those using wsh ssh) are saved to the internal connections.json config
- the connections.json config will be used to help pre-populate the dropdown list
- adds an item to the dropdown to edit the connections config in an ephemeral block


chottuthejimmy:
The onboarding is failing, here's the [rec](https://app.dashcam.io/replay/67510c2c4ebdbde81a8bc98a?share=qCJtUbvvjCpc9OIajK6rg)
Getting a NULL error

PR #1428, opened 2024-12-07
Retry the macOS builds in case of notarize failures

esimkowitz:
The notarize operation is flaky so I'm wrapping it in a retry


esimkowitz:
Retry appears to be working, gonna merge https://github.com/wavetermdev/waveterm/actions/runs/12208580986

PR #1582, opened 2024-12-19
Deleting current workspace switches to another if it can instead of closing

Jalileh:
```typescript
        if (workspaceList.length > 1 && ww?.workspaceId == workspaceId){
            
            const workspaceIdindex : any =  workspaceList.findIndex(wse => wse.workspaceid === workspaceId);
                
            const moveIndex = (workspaceIdindex > 1) ? workspaceIdindex - 1 : workspaceIdindex + 1;   
            const PrevNextWorkspace : any = workspaceList[moveIndex]
            
            await ww.switchWorkspace(PrevNextWorkspace.workspaceid);
            relaunchBrowserWindows(); // the blank window bug from 0.10.1 OCCASIONALLY occurs if this is not here,  ( issue #1567 )
        }
        
        await WorkspaceService.DeleteWorkspace(workspaceId);
```


esimkowitz:
Thank you for this submission! Please take a look at the CLA and sign it when you can. We cannot accept your submission without this. Let me know if you have any questions.

emain/emain-window.ts
```diff
@@ -699,9 +699,25 @@ ipcMain.on("delete-workspace", (event, workspaceId) => {
...
+        if (workspaceList.length > 1 && ww?.workspaceId == workspaceId){
+            
+            const workspaceIdindex =  workspaceList.findIndex(wse => wse.workspaceid === workspaceId);
+                
+            const moveIndex = (workspaceIdindex > 0) ? workspaceIdindex - 1 : workspaceIdindex + 1;   
+            const prevNextWorkspace = workspaceList[moveIndex]
+            
+            await ww.switchWorkspace(prevNextWorkspace.workspaceid);
+			// this does it faster without restarting the window
+            ww.activeTabView.webContents.reloadIgnoringCache(); // the blank window bug from 0.10.1 OCCASIONALLY occurs if this is not here,  ( issue #1567 )
```
esimkowitz, line 712: Is this still happening in the latest release?

emain/emain-window.ts
```diff
@@ -699,9 +699,25 @@ ipcMain.on("delete-workspace", (event, workspaceId) => {
             console.log("user cancelled workspace delete", workspaceId, ww?.waveWindowId);
             return;
         }
+        
+        if (workspaceList.length > 1 && ww?.workspaceId == workspaceId){
```
esimkowitz, line 703: This should be combined with the if statement on line 720. The behavior should be:

```typescript
if (ww?.workspaceId == workspaceId) {
    if (workspaceList.length > 1){
        // switch workspace
        ...
    } else {
        // delete workspace
        ...
    }
}
```

esimkowitz (changes requested):
Requesting a change and asking for clarity on the 0.10.1 bug.

I like this as the new default behavior, thank you very much for contributing this!

emain/emain-window.ts
esimkowitz: Also, while it's extremely unlikely, workspaceList returned by `WorkspaceService.ListWorkspaces` could be null so we should probably have `workspaceList?.length`

emain/emain-window.ts
Jalileh: image
been doing some testing on
image

I synced and pulled locally and thereafter tested it again without the  ww.activeTabView.webContents.reloadIgnoringCache();, still happens from time to time without that line.

esimkowitz: is this specifically in your branch where you notice this happening or do you notice this in the main version of the app too?

emain/emain-window.ts
```diff
@@ -699,11 +699,24 @@ ipcMain.on("delete-workspace", (event, workspaceId) => {
...
-        console.log("delete-workspace done", workspaceId, ww?.waveWindowId);
-        if (ww?.workspaceId == workspaceId) {
-            console.log("delete-workspace closing window", workspaceId, ww?.waveWindowId);
-            ww.destroy();
+        
+        if (ww?.workspaceId == workspaceId){
+            if ( workspaceList?.length > 1 ) {
+                const workspaceIdindex =  workspaceList.findIndex(wse => wse.workspaceid === workspaceId);
+                    
+                const moveIndex = (workspaceIdindex > 0) ? workspaceIdindex - 1 : workspaceIdindex + 1;
```
esimkowitz, line 707: There's actually a potential for conflict if the next or previous workspace is already claimed by a new window. We need to add a while condition here to iterate through the list until it finds one that isn't claimed by another window. If it can't find it, then it should destroy the window. 

Additionally, I think your change introduces a race between the Go backend and the JS code regarding workspace deletion and workspace switching.

Right now, WorkspaceService.DeleteWorkspace will clean up the window object in the database if there's an attached window. If your call to switchWorkspace is not completed before the backend deletes the workspace (which is likely given the switchWorkspace call is serialized and run in an asynchronous queue), then the backend might clean up the window object, which would result in a blank screen.

This is starting to feel like something we should manage in the Go backend instead of in Electron. A call to WorkspaceService.DeleteWorkspace should determine whether there's another workspace that the window can claim. If not, then it should clean up the window object like it already does and return an empty string to Electron, which should then destroy the window. If it finds a valid workspace to switch to, it should not clean anything up and should return the id for that workspace to Electron, which can then queue a switchWorkspace action.

Sorry for the back-and-forth on this :/

emain/emain-window.ts
Jalileh: As of right now the only difference with the main version and my branch implementation is literally this, either my use of  the existing code is wrong or i've just found another way to cause the bug when switching workspaces. I noticed that and found a way to avoid it since I have absolutely no clue what's causing it.

esimkowitz: Left another comment, I think I know why this is happening but the only way to be sure it won't is to do the logic on the backend

PR #1604, opened 2024-12-23
fix: enhance background URL processing to support Windows file paths

SamimAB:
In processBackgroundUrls, made two changes -
1. replace backslashes with "\\\\", so css-tree reads it properly
2. added a check for windows like file paths `/^[a-zA-Z]:\\/.test(originalUrl)`

I tried a few other fixes but this one works the best with least changes to the codebase. Hopefully this is helpful.

### DIFF
--- a/frontend/app/app-bg.tsx
+++ b/frontend/app/app-bg.tsx
@@ -26,7 +26,7 @@ function processBackgroundUrls(cssText: string): string {
     }
     const attrRe = /^background(-image)?\s*:\s*/i;
     cssText = cssText.replace(attrRe, "");
-    const ast = parseCSS("background: " + cssText, {
+    const ast = parseCSS("background: " + cssText.replace(/\\/g, '\\\\'), {
         context: "declaration",
     });
     let hasUnsafeUrl = false;
@@ -54,7 +54,7 @@ function processBackgroundUrls(cssText: string): string {
                 return;
             }
             // allow absolute paths
-            if (originalUrl.startsWith("/") || originalUrl.startsWith("~/")) {
+            if (originalUrl.startsWith("/") || originalUrl.startsWith("~/") || /^[a-zA-Z]:\\/.test(originalUrl)) {
                 const newUrl = encodeFileURL(originalUrl);
                 node.value = newUrl;
                 return;


SamimAB:
@esimkowitz the testdriver ai onboarding failed. does it require any actions from my side?

esimkowitz:
@ianjennings @chottuthejimmy could one of you take a look? Doesn’t seem like the job is getting queued

chottuthejimmy:
> @ianjennings @chottuthejimmy could one of you take a look? Doesn’t seem like the job is getting queued

Sure give a moment

chottuthejimmy:
@esimkowitz yes it wasn't triggered, I see nothing at our backend. 
While I look into it could re-run the job for me please

chottuthejimmy:
It is working fine on the main branch

esimkowitz:
@chottuthejimmy okay queued to rerun. Yeah I've noticed that the command to queue the job to your service fails intermittently. It seems most of the recent failures have been because of this.

chottuthejimmy:
Yeah, I mean I am trying to figure it out. 
One, the main branch has been scheduled to run everyday at 9 pm, so that gives us a benchmark to evaluate and make sure everything's fine, which atm seems ok. 
Two, the tests will automatically fail if you merge / close the PR before the test is finished, ya ik you know it, just wanted to make a second point 😅

That said will get something to you before tmr. 
Merry Christmas and thanks for your service 🫡

ianjennings:
> @ianjennings @chottuthejimmy could one of you take a look? Doesn’t seem like the job is getting queued

image

It seems like the GitHub secret key is not being set during the run. 

According to [this post](https://github.com/orgs/community/discussions/27026) it depends who triggers the run. @SamimAB may not have permissions to use GitHub secrets. Try adding them to the org?

We can definitely make our errors more apparent, will make a note of that.

esimkowitz:
Let's move this discussion to #1621

sawka:
hi, really appreciate that you brought this issue to our attention!  i ended up just making change to the `wsh setbg` code to properly escape windows filepaths at the time when we set the metadata.

https://github.com/wavetermdev/waveterm/pull/1636

so now you should be able to do something like this in powershell (note if using bash or a unix shell you'd also need to escape the backslashes in the shell as well):
```
wsh setbg c:\foo\bar.jpg
```

the problem with blanket replacing \ => \\ everywhere in the CSS is it could have unintended consequences.  it would also affect urls that were already properly encoded with backslashes.

SamimAB:
@sawka makes sense. appreciate it. will check it out.

PR #1614, opened 2024-12-24
Fix Term Widget not registering pointer events

### LINKED ISSUE #1357: [Feature]: Please add hyperlink support for terminal blocks
### Feature description

I am migrating from Warp and have noticed a missing feature in Wave. Seeing the output of a test run which may contain a stack trace, I have become accustomed to being able to cmd-click a source code reference and have that open up in my ide.

In the snapshot shown, using Warp, I would be able to cmd-click any source code reference, eg:

> /Users/plastikfan/dev/github/snivilised/pants/worker-pool-func-manifold_test.go:155
> /Users/plastikfan/dev/github/snivilised/pants/worker-pool-func-manifold.go:108
> /Users/plastikfan/dev/github/snivilised/pants/worker-pool-func-manifold.go:116

etc...

The same goes for web links that may appear in console output, they should also be cmd-clickable in a simlar fashion.

### Implementation Suggestion

_No response_

### Anything else?

WAVE-missing-hyperlinks

esimkowitz:
This is a bit janky. The problem is that we were placing the `xterm-viewport` div, which contains the scroll observer for the xterm contents, at a higher z-index than the xterm contents, meaning that the contents couldn't register any pointer events. If we don't put a z-index, though, the scroll bar can't accept pointer events. To get around this, I've added two observer divs, which control whether the contents or the viewport have pointer event priority. The first div, the `term-scrollbar-show-observer`, sits above where the scrollbar will be rendered. When the user hovers over it, it will cause the viewport div to move to a z-index above the contents. It will also enable a second div, the `term-scrollbar-hide-observer`, which sits above the viewport and the term contents, but not blocking the scrollbar. When the user hovers over this div (indicating their mouse has left the scrollbar), the viewport div is moved back to its original z-index and the hide observer is set to `display: none`. This gives pointer event priority back to the contents div.

This resolves an issue where the user could not click links in the terminal output.

Resolves #1357

### DIFF
--- a/frontend/app/view/term/term.tsx
+++ b/frontend/app/view/term/term.tsx
@@ -851,7 +851,6 @@ const TerminalView = ({ blockId, model }: TerminalViewProps) => {
     const scrollbarHideObserverRef = React.useRef<HTMLDivElement>(null);
     const onScrollbarShowObserver = React.useCallback(() => {
         const termViewport = viewRef.current.getElementsByClassName("xterm-viewport")[0] as HTMLDivElement;
-        console.log("scrollbar show observer", termViewport);
         termViewport.style.zIndex = "var(--zindex-xterm-viewport-overlay)";
         scrollbarHideObserverRef.current.style.display = "block";
     }, []);
--- a/frontend/app/view/term/term.tsx
+++ b/frontend/app/view/term/term.tsx
@@ -857,7 +857,6 @@ const TerminalView = ({ blockId, model }: TerminalViewProps) => {
     }, []);
     const onScrollbarHideObserver = React.useCallback(() => {
         const termViewport = viewRef.current.getElementsByClassName("xterm-viewport")[0] as HTMLDivElement;
-        console.log("scrollbar hide observer", termViewport);
         termViewport.style.zIndex = "auto";
         scrollbarHideObserverRef.current.style.display = "none";
     }, []);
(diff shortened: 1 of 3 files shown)


frontend/app/view/term/term.tsx
```diff
@@ -848,7 +848,21 @@ const TerminalView = ({ blockId, model }: TerminalViewProps) => {
...
+    const scrollbarHideObserverRef = React.useRef<HTMLDivElement>(null);
+    const onScrollbarShowObserver = React.useCallback(() => {
+        const termViewport = viewRef.current.getElementsByClassName("xterm-viewport")[0] as HTMLDivElement;
+        console.log("scrollbar show observer", termViewport);
+        termViewport.style.zIndex = "var(--zindex-xterm-viewport-overlay)";
+        scrollbarHideObserverRef.current.style.display = "block";
+    }, []);
+    const onScrollbarHideObserver = React.useCallback(() => {
+        const termViewport = viewRef.current.getElementsByClassName("xterm-viewport")[0] as HTMLDivElement;
+        console.log("scrollbar hide observer", termViewport);
```
esimkowitz, line 860:
```suggestion
```

frontend/app/view/term/term.tsx
```diff
@@ -848,7 +848,21 @@ const TerminalView = ({ blockId, model }: TerminalViewProps) => {
         termModeRef.current = termMode;
     }, [termMode]);
 
-    let stickerConfig = {
+    const scrollbarHideObserverRef = React.useRef<HTMLDivElement>(null);
+    const onScrollbarShowObserver = React.useCallback(() => {
+        const termViewport = viewRef.current.getElementsByClassName("xterm-viewport")[0] as HTMLDivElement;
+        console.log("scrollbar show observer", termViewport);
```
esimkowitz, line 854:
```suggestion
```

PR #1622, opened 2024-12-24
Allow TestDriver to run on outside contributions

### LINKED ISSUE #1621: TestDriver pipeline fails for non-org members
The TestDriver pipeline fails if the user who created a PR does not have GitHub Secrets permissions. This makes sense, but it also limits our ability to test external contributions before merging them.

esimkowitz:
Discussed in #1621, the current setup for the TestDriver workflow doesn't allow it to run for PRs from outside contributors, since GitHub restricts secrets access for external PRs using the `pull_request` trigger.

This PR splits the TestDriver workflow into two separate workflows, a `pull_request`-triggered one which builds the artifact, and a `workflow_run`-triggered one which can access our secrets to upload the job to TestDriver.

resolves #1621

### DIFF
--- a/.github/workflows/merge-gatekeeper.yml
+++ b/.github/workflows/merge-gatekeeper.yml
@@ -23,4 +23,4 @@ jobs:
               uses: upsidr/merge-gatekeeper@v1
               with:
                   token: ${{ secrets.GITHUB_TOKEN }}
-                  ignored: Test Onboarding, Analyze (go), Analyze (javascript-typescript), License Compliance, CodeRabbit
+                  ignored: Run TestDriver.ai, Analyze (go), Analyze (javascript-typescript), License Compliance, CodeRabbit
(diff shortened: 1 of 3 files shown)


chottuthejimmy:
Hey @esimkowitz I think as Ian said, gh sometimes passes empty secrets. 
Hope this [code](https://github.com/wavetermdev/waveterm/blob/68d6d4901dedebe00387079bc7b32c6465383d33/.github/workflows/testdriver.yml#L30C1-L32C24) doesn't cause that

esimkowitz:
It shouldn't, though is that token really necessary? The artifacts should be public. If you have your own token on your end that would be better to use so we don't risk leaking secrets

chottuthejimmy:
Ofc, always open to learn. 
Do you want to sync or just send it to me ?

PR #1623, opened 2024-12-25
Deleting current workspace switches to another instead of closing [ backend implementation ]

Jalileh:
I did not mean to close the previous pr, anyway i tried to implement what you suggested, the backend now does most of it 
and DeleteWorkspace will return an unclaimed id and avoid closing the window.

```go 
        const moveToNewWorkspace = await WorkspaceService.DeleteWorkspace(workspaceId) 
        console.log("delete-workspace done", workspaceId, ww?.waveWindowId);
        if (ww?.workspaceId == workspaceId){
            if ( workspaceList?.length > 1 ) {
                   await ww.switchWorkspace(moveToNewWorkspace)
            } else {
                    console.log("delete-workspace closing window", workspaceId, ww?.waveWindowId);
                    ww.destroy();
            }
        }
    });
```

unknown_2024 12 26-17 05
unknown_2024 12 26-17 06


pkg/wcore/workspace.go
```diff
@@ -114,40 +115,58 @@ func UpdateWorkspace(ctx context.Context, workspaceId string, name string, icon
...
 		if err != nil {
-			return false, fmt.Errorf("error closing tab: %w", err)
+			return false, "", fmt.Errorf("error closing tab: %w", err)
 		}
 	}
 	windowId, err := wstore.DBFindWindowForWorkspaceId(ctx, workspaceId)
 	err = wstore.DBDelete(ctx, waveobj.OType_Workspace, workspaceId)
 	if err != nil {
-		return false, fmt.Errorf("error deleting workspace: %w", err)
+		return false, "", fmt.Errorf("error deleting workspace: %w", err)
```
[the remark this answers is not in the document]
Jalileh: NOT ME

emain/emain-window.ts
```diff
@@ -699,11 +701,16 @@ ipcMain.on("delete-workspace", (event, workspaceId) => {
...
         }
-        await WorkspaceService.DeleteWorkspace(workspaceId);
+        
+        const moveToNewWorkspace = await WorkspaceService.DeleteWorkspace(workspaceId) 
         console.log("delete-workspace done", workspaceId, ww?.waveWindowId);
-        if (ww?.workspaceId == workspaceId) {
-            console.log("delete-workspace closing window", workspaceId, ww?.waveWindowId);
-            ww.destroy();
+        if (ww?.workspaceId == workspaceId){
+            if ( workspaceList?.length > 1 ) {
```
esimkowitz, line 708: this should check if the new workspace id is not null, the workspaceList check should no longer be necessary since the backend is dealing with this

emain/emain-window.ts
```diff
@@ -699,11 +701,16 @@ ipcMain.on("delete-workspace", (event, workspaceId) => {
             console.log("user cancelled workspace delete", workspaceId, ww?.waveWindowId);
```
esimkowitz, line 701: It won't let me add a comment above this line, but you should also remove lines 690-692 and edit the message in the Delete Workspace message box to remove the part mentioning that the window will be closed, since we're changing this behavior.

emain/emain-window.ts
```diff
@@ -301,9 +303,9 @@ export class WaveBrowserWindow extends BaseWindow {
 
         // If the workspace is already owned by a window, then we can just call SwitchWorkspace without first prompting the user, since it'll just focus to the other window.
         const workspaceList = await WorkspaceService.ListWorkspaces();
-        if (!workspaceList?.find((wse) => wse.workspaceid === workspaceId)?.windowid) {
-            const curWorkspace = await WorkspaceService.GetWorkspace(this.workspaceId);
-            if (isNonEmptyUnsavedWorkspace(curWorkspace)) {
+        if (!workspaceList?.find((wse) => wse.workspaceid === workspaceId)?.windowid) {		
+            const curWorkspace = await WorkspaceService.GetWorkspace(this.workspaceId);  /* potential bug here, curWorkspace undefined after deleting previous workspace */
```
esimkowitz, line 307: Hmm, this bug seems strange, you're still noticing this after your backend change?

pkg/wcore/workspace.go
```diff
@@ -114,40 +115,58 @@ func UpdateWorkspace(ctx context.Context, workspaceId string, name string, icon
...
+	})
+
 	if windowId != "" {
-		err = CloseWindow(ctx, windowId, false)
+		workspaces, err := ListWorkspaces(ctx)
+		if err != nil {
+			return false, "", fmt.Errorf("error retrieving workspaceList: %w", err)
+		}
+		UnclaimedWorkspace := ""
+		for _, ws := range workspaces {
```
esimkowitz, line 153: I liked your old logic which would attempt to find the workspace at index closest to the one that was deleted, could you incorporate that here?

I think you can do this in one pass. Create three variables, `curWsIndex`, `freeWsIndexBeforeCurrent` and `freeWsIndexAfterCurrent`. Then, you iterate through the workspace list, setting `freeWsIndexBeforeCurrent` to the index in the workspace list where you find a free workspace (not claimed by a window). Once you find the current index, you set `curWsIndex` and from there you continue iterating until you find another free workspace, where you set `freeWsIndexAfterCurrent`. Then, you just take whichever one of the two indices is closer to `curWsIndex` (or whichever isn't null).

esimkowitz (changes requested):
Looking good! Left some suggestions and questions

.vscode/launch.json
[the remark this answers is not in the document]
Jalileh: yea didnt mean to sorry, it was my way of debugging golang. Is the debugging experience this miserable in the language? 

everytime I want to disable the ctx timeouts im doing this
image
which in turn causes this because my stupid vim config formats the second i open the buffers with cdo
Screenshot 2024-12-26 172101

emain/emain-window.ts
Jalileh: image
image

The bug occurs both in the backend and in electron, because whenever we delete the current workspace we still try to access the 'currently deleted workspace' from the window object.

pkg/wcore/workspace.go
Jalileh: sure i can implement that if we can ignore this return error, because im relying on getting any claimable workspace to not upset this code because deleteworkspace also runs in switchworkspace for some reason i dont know
image

.vscode/launch.json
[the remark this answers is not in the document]
esimkowitz: Heh, yeah it's not great. I wanna put together a better debugging env for our app, but I think because of the way everything needs to spawn in order, it's gonna need to be kind of bespoke. 

I also want to build out a better way of mocking the frontend<->backend exchanges so we can develop them in a bit more isolation, we've just been so consumed with features that I haven't had time to focus on non-production devops. Hopefully we can make it better in the new year.

pkg/wcore/workspace.go
esimkowitz: Ah it runs in switch workspace to clean up unsaved workspaces. For saved ones, delete without a force flag is a no-op

emain/emain-window.ts
```diff
@@ -18,6 +18,8 @@ import { delay, ensureBoundsAreVisible, waveKeyToElectronKey } from "./emain-uti
 import { log } from "./log";
 import { getElectronAppBasePath, unamePlatform } from "./platform";
 import { updater } from "./updater";
+import { time } from "console";
+import { timestamp } from "rxjs";
```
esimkowitz, line 22: why are these needed?

Jalileh: lint moment, removing that...

pkg/wcore/workspace.go
Jalileh: how do I actually retrieve the error string in the error data structure?  if I can get the string to reliably know if 'not found' happened I wont have to limit the claimworkspace implementation, I only saw msg inside of it but couldn't access it, sorry I'm new to go.

esimkowitz: I think you can return empty string if nothing is deleted. This should never happen if you call deleteworkspace from the workspace service since that passes the force flag as true to the wcore implementation

Jalileh: OK ill implement the old logic tomorrow and notify you, ty!

pkg/wcore/workspace.go
```diff
@@ -114,40 +115,76 @@ func UpdateWorkspace(ctx context.Context, workspaceId string, name string, icon
...
+			err = CloseWindow(ctx, windowId, false)
+		}
+
 		if err != nil {
-			return false, fmt.Errorf("error closing window: %w", err)
+			return false, "", fmt.Errorf("error closing window: %w", err)
 		}
 	}
-	return true, nil
+	return true, "", nil
```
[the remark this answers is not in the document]
Jalileh: Deleting workspaces with/without unclaimed workspaces before/after TESTED ✔
Force deletion vs. protected named workspaces  ???  
Window reassignment logic  TESTED WORKS WONDERFULLY ✔
Tab cleanup during workspace deletion  TESTED ✔
Error handling for various failure scenarios TESTED/DEBUGGED x4^1000 ✔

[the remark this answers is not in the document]
Jalileh: I dont think I want to talk to you anymore

esimkowitz:
Just tested, it works great! I've submitted a small formatting fix for the TypeScript files, I'm guessing you don't have Prettier set up for your Neovim. I'm going to sign off on this, feel free to merge it when you're ready 😃 

Thank you so much for your contribution!!

esimkowitz (approved):

Issue #1613, opened 2024-12-23
[Feature]: Customize default window size and zoom


esimkowitz:
### Feature description

Add a configuration to override the default new window size and zoom so users with larger/smaller monitors can get the best experience.

### Implementation Suggestion

Add `window:dimensions` setting with a string value `{X}x{Y}` where the units are pixels. This can be applied in `emain-window.ts` when the Electron BaseWindow object gets initialized.

Add `window:zoom` setting with a float value that defaults to 1. This can be applied in `emain-tabview.ts`, where the WebContentsView gets initialized. This can be overridden by different tabs too.

zaidnadaff:
Hi, can I work on this?

esimkowitz:
Of course! Assigning to you

esimkowitz:
Updated the ask for `window:zoom`, to match what [Electron expects](https://www.electronjs.org/docs/latest/api/web-contents#contentssetzoomfactorfactor), this should probably be a float value instead of an integer

zaidnadaff:
Hey @esimkowitz I'm sorry for asking as I am new to this.

For the first part of making sure that window size gets overridden, a combination of ScaleFactor and relative views like percentage can be used? ScaleFactor makes sure that the physical and logical pixels of devices with different resolutions maintain a uniform view. 

For the zoom, I did not quite get the requirement is.

Also, I noticed another error, when zooming out, when I was pressing ctrl + '-' for a continuously, the app crashed. 

![Screenshot from 2024-12-27 10-17-04](https://github.com/user-attachments/assets/918f5b8f-1a6b-4759-a548-5398d2696bdc)

this was fixed when I replaced this code 

`                if (wc.getZoomFactor() <= 0.2) {
                    return;
                }
                wc.setZoomFactor(wc.getZoomFactor() - 0.2);` 

with this

`                wc.setZoomFactor(Math.max(0.2, wc.getZoomFactor() - 0.2));
`

esimkowitz:
I think for the default zoom factor, it's just whatever the user defines is the default when a new window gets created and the value we set it back to when they use the Cmd:0 keyboard shortcut

PR #1627, opened 2024-12-25
fix: pass missing config to setTimeout in getSpareTab

BQXBQX:
- Fixed an issue where `fullConfig` was not passed to `setTimeout` in `getSpareTab`.
- Wrapped `ensureHotSpareTab` with an arrow function to ensure `fullConfig` is properly passed.

**When I use wave as my terminal, sometimes the setting does not work. I read the source code and found this issue**

### DIFF
--- a/emain/emain-tabview.ts
+++ b/emain/emain-tabview.ts
@@ -263,7 +263,7 @@ export function ensureHotSpareTab(fullConfig: FullConfigType) {
 }
 
 export function getSpareTab(fullConfig: FullConfigType): WaveTabView {
-    setTimeout(ensureHotSpareTab, 500);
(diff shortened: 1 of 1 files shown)


esimkowitz:
What settings specifically were not getting applied, the default background? This looks like the only one that's actually supplied by fullConfig

BQXBQX:
Yes, when I create a new tabview, the fullConfig in the constructor of tabview is undefined and the background color will have problems, I have also tried to solve https://github.com/wavetermdev/waveterm/issues/1613. The zoom part of this issue may also report errors due to this.

esimkowitz:
Looks reasonable to me, will merge

PR #1634, opened 2024-12-27
Single Session Mode

oneirocosm:
Allows connections to spawn multiple sessions from only one remote session. Useful specifically in situations where sshd `ForceCommand` is being used for 2-factor authentication. The main idea is that a single connection session is made with a command containing the wsh connserver. Then, the connserver is able to spawn local sessions that can be routed through the connserver instead of through individual shells.


oneirocosm:
The code has changed significantly since this draft was written, so it is no longer relevant.

Issue #1608, opened 2024-12-23
[Feature]: Exportable configuration


Printoiid:
### Feature description

It would be a nice to have to export my configuration and associated folders into a singular folder so that I can back up any backgrounds, themes, settings, and presets to GitHub so I can use it in the future if I need to reinstall my operating system, without worrying about anything being lot

### Implementation Suggestion

N/a, just an "export" command with WSH to a file path would be great

### Anything else?

n/a

esimkowitz:
As a workaround, all your configs are stored in ~/.config/waveterm, should be easy enough to add to your dotfiles sync script.

weilirs:
Hi, can I work on this?

esimkowitz:
Sure, I'll assign to you. What's your planned implementation? I was thinking it could be cool to zip up the config dir and put it at a user-provided directory

weilirs:
> ### Feature description
> It would be a nice to have to export my configuration and associated folders into a singular folder so that I can back up any backgrounds, themes, settings, and presets to GitHub so I can use it in the future if I need to reinstall my operating system, without worrying about anything being lot
> 
> ### Implementation Suggestion
> N/a, just an "export" command with WSH to a file path would be great

Yeah, I think this would be cool. I'll go with this approach then.

weilirs:
If the PR looks good I'll update the documentation for this command.

esimkowitz:
The PR looks good to me, made one small suggestion based on the CodeRabbit comment, but otherwise I think it's good to go!

Lizardbutt:
Wouldn't it be a good idea to associate this Feature with the PR itself?

Lizardbutt:
Alright, now we need another to import configs from a path. lmfao

esimkowitz:
The logic there is a bit trickier, need to merge with existing configs and determine the override logic there

PR #1641, opened 2024-12-27
feat: wsh `export-config` command

weilirs:
feat #1608 
`export-config [output-path]` to export config files into a zip archive.

### DIFF
--- a/cmd/wsh/cmd/wshcmd-export-config.go
+++ b/cmd/wsh/cmd/wshcmd-export-config.go
@@ -1,0 +1,88 @@
+package cmd
+
+import (
+	"archive/zip"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	var exportConfigCmd = &cobra.Command{
+		Use:   "export-config [output-path]",
+		Short: "export Wave Terminal configuration",
+		Long:  "Export Wave Terminal configuration files into a zip archive",
+		RunE:  runExportConfig,
+	}
+	rootCmd.AddCommand(exportConfigCmd)
+}
+
+func runExportConfig(cmd *cobra.Command, args []string) error {
+	if len(args) != 1 {
+		return fmt.Errorf("export-config requires an output path")
+	}
+
+	outputPath := args[0]
+	if !strings.HasSuffix(outputPath, ".zip") {
+		outputPath += ".zip"
+	}
+
+	configDir := getWaveConfigDir()
+	if err := zipDir(configDir, outputPath); err != nil {
+		return fmt.Errorf("export-config failed: %v", err)
+	}
+
+	fmt.Printf("Configuration exported to %s\n", outputPath)
+	return nil
+}
+
+func zipDir(sourceDir, zipPath string) error {
+	zipFile, err := os.Create(zipPath)
+	if err != nil {
+		return err
+	}
+	defer zipFile.Close()
+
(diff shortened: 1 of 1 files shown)


cmd/wsh/cmd/wshcmd-export-config.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	configDir := getWaveConfigDir()
+	if err := zipDir(configDir, outputPath); err != nil {
+		return fmt.Errorf("export-config failed: %v", err)
+	}
+
+	fmt.Printf("Configuration exported to %s\n", outputPath)
+	return nil
+}
```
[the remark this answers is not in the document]
esimkowitz: @weilirs this shouldn't ever happen but it's probably good for posterity. Could you check that the config dir path is not empty before walking?

esimkowitz (changes requested):

weilirs:
I added a check for the existence of config directory, please take a look

esimkowitz:
Looks good to me! Can you add documentation to the `wsh-reference.mdx` file for your new command?

weilirs:
> Looks good to me! Can you add documentation to the `wsh-reference.mdx` file for your new command?

I added to the documentation and I modified the command from `export-config` to `exportconfig` for consistency.

cmd/wsh/cmd/wshcmd-export-config.go
```diff
@@ -0,0 +1,93 @@
...
+		if err != nil {
+			return err
+		}
+
+		_, err = io.Copy(writer, file)
+		return err
+	})
+}
+
+func getWaveConfigDir() string {
```
esimkowitz, line 87: This is not reliable as the user may have a different config directory defined, such as on dev where it's `~/.config/waveterm-dev`. You should use `wavebase.GetWaveConfigDir` instead.

esimkowitz: However, because this is called from the wsh side, you should just call the `wshclient.PathCommand` function, since that will get you whatever the value is in wavesrv

docs/docs/wsh-reference.mdx
```diff
@@ -153,6 +153,16 @@ wsh editconfig presets/ai.json
...
 
+## exportconfig
+
+You can export your Wave's config files into a zip archive using this command.
+
+```
+wsh exportconfig [output-path]
+```
+
+---
```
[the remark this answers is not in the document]
esimkowitz: good suggestion @weilirs

cmd/wsh/cmd/wshcmd-export-config.go
esimkowitz: actually, this won't work at all if you're on a remote machine. The logic for zipping the config should be moved to wavesrv and called through the wshserver mechanism, you can look at the other commands to see how we do this. Then, if the user is invoking this command from a remote machine, we should copy the zipped archive to the output path on the remote machine

weilirs: To copy the zipped archive to the output path on the remote machine, should I use the `filestore.WFS.WriteFile` or something else? I am using `filestore.WFS.WriteFile` but `ls` doesn't show anything on the remote machine. thanks!

esimkowitz: I'm working on a more flexible filesystem traversal mechanism right now, targeting end of this week. Once this is implemented, I'll let you know so you can integrate it here. For now, let's hold on this PR

sawka:
closed due to age

PR #1646, opened 2024-12-28
Prefui - Preferences User Interface 

Jalileh:
_**Why am I going to work on this?**_
 Because I want to implement a bunch of customization options and need them not to clog the right click list.

**_How will it work?_**
I'll have a Preferences option in the ... list, whenever pressed a Preferences tab will open or ill just put it to the right under sysinfo.

**_What will be on there_**
 - Option to set fonts for shell e.g ( FiraCodeNerdFont-Retina.ttf ) 💯 
 - Option to set start directory of shell  💯 
 - Option to set default type of shell  💯 
 - Option to set opacity and other stuff maybe
 - Maybe keybinds? 

I'll add more stuff but above is what I'm aiming for at the moment and of course good implementation


esimkowitz:
Looks like this PR is empty, is it intended for review? If not, could you convert it to a draft?

Jalileh:
```Go

/*
GUIName: for yaml and alternative db id, mostly for electron
*/
type prefItem struct {
	Name    string            `json:"name"`
	GUIName string            `json:"guiname"`
	Config  map[string]string `json:"config"`
}
```

This looks OK so far?

3 files of go and I doubt I'll need more for this,
I'm unsure if I should integrate the main data structure into wtype.go or just do it in one of the files myself.
Any changes I should do before going anywhere with this?

esimkowitz:
sorry, could you elaborate on your goals for this PR? Why does this require a new preferences system?

Jalileh:
> Most of these features are already supported by our current [config system](https://docs.waveterm.dev/config) (`term:localshellpath`, `cmd:cwd`, `term:fontfamily`). Was there a reason these weren't working for your use case? Maybe there's a short-term documentation gap we can address here.
> 
> I guess I'm not really sure where this fits into our vision for customization. We've got some bigger plans to overhaul the config system, but it keeps getting punted because we had other things that were more important. The main things I want to improve are how presets override existing configs (right now, they are kinda handled differently for different use cases, which is really annoying, and they're not dynamic, they only get read when you select a preset to apply and then you need to select it again to get the latest values).
> 
> We do desperately need to improve the config story and add things like a json schema and a UI for modifying settings, we just haven't gotten around to it yet. Hopefully we can take some time to address this in the next couple releases.

Wow, my fault for not reading docs it's great for what's already there, i'll work on the json schema and the UI if it's okay.

esimkowitz:
Just DM'ed you on Discord, let's discuss further :)

PR #1663, opened 2025-01-01
fix: remove context from wsl connection struct

oneirocosm:
This fixes a WSL bug where disconnecting the controller does not allow the controller to restart until the app reboots.


sawka:
might be able to add an actual timeout to some of these... not the connserver, but for the other utilities -- like shell detection or wsh path, etc. they could just have like a 5s timeout.

Issue #1625, opened 2024-12-25
[Bug]: When maximizing a non-focused block, the block border is set incorrectly.


ZLATAN628:
### Current Behavior

When I maximize a non-focused block,  the block border is set incorrectly. I tried debugging the source code to find the answer and discovered that the width and height are first assigned the correct values but are then updated back to their original values.

<img width="2560" alt="image" src="https://github.com/user-attachments/assets/0e353467-fab2-4bcf-9236-aee63a22cafb" />

### Expected Behavior

<img width="2560" alt="image" src="https://github.com/user-attachments/assets/4e815f69-a99a-4414-be0c-9dab1ba5b98d" />

### Steps To Reproduce

Clicking the maximize button on the non-focused block 

### Wave Version

main

### Platform

macOS

### OS Version/Distribution

macos

### Architecture

arm64

### Questionnaire

- [ ] I'm interested in fixing this myself but don't know where to start
- [X] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

esimkowitz:
I'm not able to reproduce this, could you record a video? Also how are you magnifying the block, using the context menu item or the button in the block header?

ZLATAN628:
Please watch this video. I have tried the latest release and dev versions and this problem exists in all widget.

https://github.com/user-attachments/assets/39a7b5cd-f480-4921-bc70-a3a747e1ff7e

ZLATAN628:
Hi @esimkowitz ,Please watch this video. I have tried the latest release and dev versions and this problem exists in all widget.


https://github.com/user-attachments/assets/5c510dc9-b411-4391-8d34-c612a3085b41

esimkowitz:
That's so strange, this does not reproduce on my system...

ZLATAN628:
I am not very good at front-end. I found that the event of updating the window size is triggered twice. The first one should be triggered by the focus event, so it enters the setInnerRectDebounced delayed update, and the second one is triggered by maximizing the window, but it is directly updated immediately here. So the performance is that after maximizing the window, it is normal at first, and then because of the delayed update, it is updated back to the window size before maximization.
<img width="967" alt="GoLand 2024-12-27 08 29 55" src="https://github.com/user-attachments/assets/4b724435-1681-4571-80c4-8b90776fbc8e" />


https://github.com/user-attachments/assets/4393c6e8-45d0-42b3-8df7-59dfe1ae02e7

[...]

PR #1702, opened 2025-01-10
fix issue #1625

ZLATAN628:
Created a global map cache for each block’s timeout event to address issue #1625.


sawka:
closed due to age

Issue #1694, opened 2025-01-08
[Bug]: If a folder is soft link, it will be treated as a file and not a folder


abgox:
### Current Behavior

- If a folder is soft link, it will be treated as a file and not a folder. And associated icons, types, etc. are all wrong.

![recording](https://github.com/user-attachments/assets/cda280f8-afb6-4d98-a7b6-0b863c4fbb7e)

### Expected Behavior

They should be correctly recognized as a folder.

### Steps To Reproduce

When you use `File Preview` to browse a file directory, this error occurs whenever a folder in the directory is soft link.

### Wave Version

0.10.4

### Platform

Windows

### OS Version/Distribution

Windows 11

### Architecture

x64

### Questionnaire

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

Jalileh:
hey are you using cmd's mklink to create symbolic links by any chance?

abgox:
- I use  the pwsh command `New-Item -ItemType SymbolicLink xxx` to create the soft link.
- I think pwsh and cmd may work the same way.

Jalileh:
> - I use  the pwsh command `New-Item -ItemType SymbolicLink xxx` to create the soft link.
> - I think pwsh and cmd may work the same way.


Noticed more of my files/folders were also bugged and I don't use CMD.

Doesn't matter if we dont use cmd, other scripts/programs will eventually use it like the .config folder.

Golang os package literally wont return the bits for CMD made symlinks from fileInfo.Mode all it returns is read perm bits.

esimkowitz:
closing as we're about to release 0.11. Thank you @Jalileh for fixing this!!

esimkowitz:
@abgox what utility do you use to make your symlinks?

Jalileh:
@esimkowitz 

Yeah sorry, the bug here is in the dot of the symlink folder name ( .config ) 
PR bot ironically made it less robust at the cost of my over-engineered approach

```go
 	isFileSymlink := func(filepath string) bool {
		Length := len(filepath) - 1
		maxFileDotExt := 4 // should cover most file extensions
		for i := Length; i >= (Length - maxFileDotExt); i-- {
			if filepath[i] == '.' {
				return true
			}
		}
		return false
	}
```
[...]

PR #1710, opened 2025-01-10
fileMode-fix for issue #1694 Windows bug

Jalileh:
windows symlinks are strange and they are not treated as directories in go, so using just fileInfo.Dir isn't enough.
Now treated as directories.

PR branch 
image
main branch 
image

bug issue link:  https://github.com/wavetermdev/waveterm/issues/1694


Jalileh:
Hey for some reason golang fs package uses 0x8000000 which is not the real FILE_ATTRIBUTE value of windows,
unless that value is for cross-compability my changes may screw with linux 
image

esimkowitz:
I tested this on Linux and it doesn't seem to affect anything so I think we're good there. What tool were you using to make symlinks on Windows? When I use `junction`, I'm still seeing it show up as a file:
image

Jalileh:
> I tested this on Linux and it doesn't seem to affect anything so I think we're good there. What tool were you using to make symlinks on Windows? When I use `junction`, I'm still seeing it show up as a file:
> image
> 

mklink

esimkowitz:
seems like it should also support `junction` for completeness
https://learn.microsoft.com/en-us/sysinternals/downloads/junction

Jalileh:
> seems like it should also support `junction` for completeness, not sure why they operate differently but such is Windows 😅 
> https://learn.microsoft.com/en-us/sysinternals/downloads/junction

Hey give one more commit and I'll see what I can do.

forgot to account for file symlinks and I'm going to have to hardcore the values for that in think, might as well resolve the junction problem if I can.

Jalileh:
image
for some reason golang maps mklink /D and /J  as  different flags while a file link has the same as /D, why?

Jalileh:
> image for some reason golang maps mklink /D and /J as different flags while a file link has the same as /D, why?

image
As you can see this makes my whole PR pointless, because of the high level fake flags for windows.
To access real window flags id need to make a seperate file for windows im glad to do this if youd like and it would be much easier than rely on this compability interface.

Jalileh:
Maybe this isn't necessary? What if /D is uncommon to use? I don't know, but the junction flag wont interfere and if it's the standard then I doubt this bug will be common, i hate cmd.

esimkowitz:
Just pulled and it looks great to me, thanks for fixing this! Are you ready to merge?

PR #1728, opened 2025-01-14
Fix for test results not displaying in checks tab

chottuthejimmy:


chottuthejimmy:
Hey @esimkowitz So github workflow_run triggers run in a separate pipeline that only knows the “completed” event from another workflow’s outcome; it doesn’t automatically attach itself to the original commit or PR.
That's the reason the `TestDriver.ai Run` wasn't being displayed. 
There are a couple of ways to fix this. 
- can call GitHub’s checks REST API (via actions/github-script or any other method) to create/update a check run that references the commit associated with the pull request. It effectively “attaches” the `workflow_run` job to the PR’s commit checks.
- or can merge the `build` and `run` into a single workflow. 

in this PR I have tried to implement the former, unfortunately I was unable to test it on my fork branch. 
But it should work.

esimkowitz:
I think those are all my comments, yeah I think the best long-term approach is to ditch the actions altogether. You can set up a server that listens to webhook events and updates the PR checks directly via the GitHub API using a GitHub App Token. I believe this is how CodeRabbit does their automated checks and we found it super painless to set up. We can check a file into our repo root that contains the TestDriver prompts and configs so you can pick it up in your system. Obviously not a short-term remedy, but food for thought as your product matures.

chottuthejimmy:
Got it.
I totally get what you are saying, really appreciate for all the inputs with nice explanation, I personally learning a ton. 

That said, I think its gonna take a while for us to build that, given the fact that we are literally a two person team and got like a million other things, we are working super hard atm. 
Building an org is hard and we are experiencing those "hard" part. 

So, if you could merge this for now and basically have a way to check it, would be great!

Hope you understand Evan :)

esimkowitz:
Of course! Yeah just some musings I had while I was waiting on the ski lift 🤣

chottuthejimmy:
I think there's a bug in my code. 
[Failed Action](https://github.com/wavetermdev/waveterm/actions/runs/12782673544/job/35632538721)
There's an error `ReferenceError: steps is not defined`

chottuthejimmy:
@esimkowitz I need a bit of clarification.
in the TestDriver Run, we have got[Artifact Url](https://github.com/wavetermdev/waveterm/actions/runs/12804172776/job/35698071321#step:3:126) we are fetching the `runid` .
But the build is happening in a different workflow right, so shouldn't we fetch the `runid` from there ?
Maybe as the run workflow is scheduled only after the completion of the build workflow, gh just takes that runid ??

PR #1863, opened 2025-01-28
reverted a broken AI refractor, improved previous method of file checking

Jalileh:
Now symlink directories starting with  .  wont break anything


sawka:
closed due to age

Issue #2063, opened 2025-03-16
[Feature]: Retrieve messages from AI block via wsh


oqoqqo:
### Feature description

I would like to be able to read the chat history from wsh. At the moment, it's easy to send messages to wsh, but not to retrieve them. (Unless I missed something?)

PR #2064, opened 2025-03-16
Retrieve messages from AI block via wsh

oqoqqo:
Related to #2063


sawka:
closed due to age (old AI block is deprecated, should target Wave AI now)

PR #2101, opened 2025-04-06
fix: add WebSocket polyfill for Electron main process

DepsCian:
Resolves ReferenceError where WebSocket was undefined in main process.
Implements dual WebSocket loading strategy with proper environment
detection and fallback mechanisms for Node.js contexts.


nahoj:
I also have this issue and would love to see this fix merged.

PR #2245, opened 2025-08-08
Fix WebSocket utilities and Add OSC 52 clipboard support

gitmpr:
- Fix WebSocket utilities timing issue:
    * Replace async import() with synchronous require() for Node.js 'ws' module
    * Prevents race condition where newWebSocket was called before import resolved
    * Fixes fallback to browser WebSocket in Node.js environment

  - Add handleOsc52Command() for set/query/clear clipboard operations
  - Register OSC 52 handler in TermWrap constructor
  - Support base64 encoding/decoding for clipboard data
  - Add error handling for clipboard access failures


zwcf5200:
The support of the remote copy function is very much needed to facilitate the support of SSH remote development.

PR #2247, opened 2025-08-09
feat: add maxtokensfield

Salnika:
**Title**
Add `maxtokensfield` config to support custom max token param names

**Description**
Some models do not support the `max_tokens` parameter and require a different field name (e.g., `max_completion_tokens`).
Currently, this causes errors such as:

```json
{
  "error": {
    "message": "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.",
    "type": "invalid_request_error",
    "param": "max_tokens",
    "code": "unsupported_parameter"
  }
}
```

**Changes in this PR**

* Added a new config field:

  * `"ai:maxtokensfield"` (string) → defines the name of the max tokens parameter to send in requests (default: `"max_tokens"`)
* Updated schema, settings types, and docs accordingly
* Added alias support for common variations (`max_tokens_param`, `max_tokens_field`, etc.)

**Example**

```json
"ai:maxtokensfield": "max_completion_tokens",
"ai:maxtokens": 1500
```

**Impact**
This change allows compatibility with models that require a different max tokens parameter name, avoiding request errors without hardcoding model-specific logic.

First time contributing here — happy to adjust anything if needed 🙏


sawka:
thanks for submitting!

for testing... can you let me know what model(s) you got the issue with -- that required a different maxtokens field name?

sawka:
ah :) ... i see, gpt-5 is a good test case.  will try to get that one working

sawka:
as of the last release the code was updated to better choose between maxtokens and maxcompletiontokens.  it should handle the main cases (o1, o3, gpt-5, etc.).  so i'm going to close this.  if we need fine grained control we can re-open, but the current implementation is simpler for now.

Issue #2427, opened 2025-10-11
[Feature]:need the colored text editing styles from v0.11.3 ported to the latest release


TimothyChen225:
### Feature description

<img width="453" height="187" alt="Image" src="https://github.com/user-attachments/assets/c573035d-34b8-4719-9117-b9029daf77a7" />
The latest version's result is as shown above. The fonts have no color and look very unappealing. After I reverted to version 0.11.3, the software soon forced an update (it silently updated to the latest version when I wasn’t paying attention). I strongly dislike this new version. Thank you very much.

sawka:
Thanks for reporting this.  A couple quick questions to help me track this down.

What platform is this (MacOS, Windows, Linux)?  If Linux, what installer are you using (rpm, deb, appimage, etc?)

Is this just affecting the "code editor" only?  For all file types, or just certain files?  Can you tell me what the name of the file you are looking at that doesn't have highlighting (e.g. ".bashrc").  The name is important because sometimes the mime-type comes from the extension or filename vs just from the content.

My guess is that the code editor is not picking up the right mime type for your file so it isn't getting the correct syntax highlighter.

slate20:
I also am getting this behavior. No syntax highlighting or any other expected "code editor" features. Appears files are just opened in the preview widget in an edit mode? I've tested with multiple file types: .py, .go, .html, .css, and so on

<img width="978" height="631" alt="Image" src="https://github.com/user-attachments/assets/147f5698-67d8-46e4-876c-a0dae300a3d6" />

PR #2459, opened 2025-10-19
Fix Monaco editor syntax highlighting in Preview block

slate20:
The Preview block's CodeEditPreview component was not passing file information to Monaco editor, preventing automatic language detection from file extensions.

Changes:
- Add fileName prop to CodeEditor component
- Pass actual file path to Monaco with fallback to random UUID

Fixes syntax highlighting for all text files opened in Preview blocks while maintaining backward compatibility with existing CodeEditor usage.

### DIFF
--- a/frontend/app/onboarding/onboarding-command.tsx
+++ b/frontend/app/onboarding/onboarding-command.tsx
@@ -1,7 +1,7 @@
 // Copyright 2025, Command Line Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-import { useCallback, useLayoutEffect, useState } from "react";
+import { useCallback, useLayoutEffect, useRef, useState } from "react";
 import { FakeBlock } from "./onboarding-layout";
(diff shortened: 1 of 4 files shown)


sawka:
@slate20 thanks for tracking this down (https://github.com/wavetermdev/waveterm/issues/2427)!  I see it now, I had a regression from this commit a month ago (Sep 15) -- https://github.com/wavetermdev/waveterm/commit/77bbf74ef95be2122db8e1ac9481a14013bf60e2

 i swapped from the filename to a UUID, which now I see lost the language detection / highlighting!

----

the issue that my commit fixed for an issue where the model was getting initialized with an empty filename (which bound the 'internal' monaco model).  that was particularly bad when *two* files were bound with empty string and then they would cross talk and overwrite each other.

----

i'll just need to verify we dont re-introduce that bug with this fix.  will work on getting this merged on monday so it will make the next patch release (should be this week).

Issue #1933, opened 2025-02-08
[Feature]: Full screen doesn't work


denisdamata:
### Feature description

Open `waveterm` directly in full screen (<kbd>F11</kbd> button) is not working:

```
waveterm --full-screen
```

esimkowitz:
This should be doable, though starting the app with `waveterm` is only currently supported on Linux

denisdamata:
> This should be doable, though starting the app with `waveterm` is only currently supported on Linux

It's not possible and there's also a bug: you can't set it to full screen with <kbd>F11</kbd> button when you're inside the Wave terminal

denisdamata:
> This should be doable, though starting the app with `waveterm` is only currently supported on Linux

It's not possible and you can't set it to full screen with <kbd>F11</kbd> button when you're inside the Wave terminal

However, I found a partial solution to the problem [ChatGPT] using the Linux window manager called `wmcrtl`. 

Create a `.sh` file called `waveterm-full-screen` in the folder where you store your scripts, make it executable  

```
chmod +x waveterm-full-screen.sh
```

and place the following code inside it

```bash
#!/bin/bash

waveterm &

# Loop for half a second until the ‘Wave’ window is open
while ! wmctrl -l | grep -q "Wave"; do
    sleep 0.5  
done

wmctrl -r "Wave" -b toggle,fullscreen
```
You can now open `waveterm` directly in full screen by running this script instead of the binary:

```
./waveterm-full-screen.sh
```

esimkowitz:
What OS are you on? F11 is only really used for fullscreen on Windows. On my Mac, it maps to "Show desktop"

denisdamata:
> What OS are you on?

If I haven't said anything, it's because it's Linux (Ubuntu 24.10).

esimkowitz:
Not a given, but thank you for confirming. I think this is certainly doable, I also did some research and at least on Windows and Linux it seems like while F11 is not default-mapped to full-screen, it's also not default-mapped to anything else so we should be able to do this. I'd prefer to handle the launch in full-screen mode as a user setting since we don't currently have a way to pass arguments when launching waveterm

Mostafa-Khairy0:
I would like to propose adding a new setting option that allows users to choose whether the application should always launch in full-screen mode or not.

This gives users more flexibility and control over how the application appears based on their workflow.

If approved, I’m happy to implement this feature and open a Pull Request. 
@esimkowitz

[...]

PR #2490, opened 2025-10-29
feat: Add fullscreen launch setting (#1933)

Mostafa-Khairy0:
New UI Setting: Launch On Fullscreen

A new option has been added to the block context menu allowing users to enable fullscreen mode automatically when Wave launches. This setting is persistent and can be toggled on a per-user basis.

How to Enable:
	1.	**Right-click on any tile/block.**
	2.	**Open the settings menu.**
	3.	**Select Launch On Fullscreen → On.**
Screenshot 2025-10-29 at 4 43 21 PM


sawka:
@Mostafa-Khairy0 thanks for the submission!  I think this context menu item would be better in the main wave menu (see emain-menu.ts) in the "View" menu rather than in the terminal context menu.

PR #2505, opened 2025-11-02
chore: add package-lock.json to .gitignore

Mostafa-Khairy0:
### Summary
This PR adds `package-lock.json` to the `.gitignore` file to prevent the npm lock file from being committed.

### Rationale
WaveTerm appears to use multiple package managers and rebuild flows (npm, pnpm, electron-builder). Ignoring `package-lock.json` avoids version lock conflicts and reduces merge noise between environments.

### Changes
- Added `package-lock.json` to `.gitignore`
- Untracked existing lock file if present

### Impact
No runtime or build impact — purely affects Git tracking behavior.


sawka:
package-lock.json MUST be committed and checked in.  this makes sure that each build gets exactly the same dependencies checked out (it is important for all projects to check in a lock file).

We only use npm right now.  AFAIK we removed all the "yarn" dependencies and never added any pnpm code.  If you see some place that is calling an alternate package manager let me know.

PR #2523, opened 2025-11-05
feat(terminal): improve input handling for interactive CLI tools

togo01:
## Summary

This PR includes three terminal input improvements that enhance the experience when using interactive CLI tools and IME input:

1. **Shift+Enter newline support**: Enable Shift+Enter to insert newlines by default
2. **Image paste support**: Allow pasting images by saving them as temporary files and pasting the file path
3. **IME duplicate input fix**: Fix duplicate text when switching input methods during composition

### Shift+Enter for newlines
Currently, pressing Shift+Enter in the terminal behaves the same as Enter, making it difficult to input multi-line commands or text in interactive CLI tools. This change enables Shift+Enter to insert newlines by default, matching common terminal emulator behavior.

### Image paste support
Interactive AI tools like Claude Code support receiving images through file paths, but Wave Terminal currently doesn't support pasting images. This change implements image paste functionality similar to iTerm2's behavior: when an image is pasted, it's saved to a temporary file and the path is pasted into the terminal.

### IME duplicate input fix
When using Chinese/Japanese/Korean IME in the terminal, switching input methods with Capslock during composition causes the composed text to be sent twice, resulting in duplicate output (e.g., "你好" becomes "你好你好"). This issue severely impacts users who frequently switch between languages.

### Shift+Enter newline (`frontend/app/view/term/term-model.ts`)
- Change default `shiftenternewline` config from `false` to `true`
- Send standard newline character (`\n`) instead of escape sequence (`\^[\n`)

### Image paste (`frontend/app/view/term/term-model.ts`, `frontend/app/view/term/termwrap.ts`)
- Add `handlePaste()` method to intercept Cmd+Shift+V paste events
- Add `handleImagePasteBlob()` to save images to `/tmp` and paste the file path
- Detect image data in clipboard using both `ClipboardEvent.clipboardData` and Clipboard API
- Support both screenshot paste and file copy scenarios
- Add 5MB size limit for pasted images
- Temporary files are created with format: `waveterm_paste_[timestamp].[ext]`

### IME duplicate input fix (`frontend/app/view/term/termwrap.ts`, `frontend/app/view/term/term-model.ts`)

**IME Composition Handling:**
- Track composition state (isComposing, composingData, etc.) in TermWrap
- Register compositionstart/update/end event listeners on xterm.js textarea
- Block all data sends during composition, only allow after compositionend
- Prevents xterm.js from sending intermediate data during compositionupdate phase

**Deduplication Logic:**
- Implement 50ms time window deduplication for both IME and paste operations
- Track first send after composition, block duplicate sends from Capslock switching
- Ensure Ctrl+Space and Fn switching work correctly (single send only)

**Edge Case Handling:**
- Add blur event handler to reset composition state on focus loss
- Add Escape key handling to cancel composition in progress

### Shift+Enter
1. Open a terminal in Wave
2. Press Shift+Enter
3. Verify that a newline is inserted instead of executing the command

### Image paste
1. Take a screenshot and copy it to clipboard (or copy an image file in Finder)
2. In a terminal running Claude Code, paste the image (Cmd+V or Cmd+Shift+V)
3. Verify that the image path appears and Claude Code recognizes it

### IME Input Testing

**IME Input:**
- [x] macOS Zhuyin IME + Capslock switching - no duplicate output ✅
- [x] macOS Zhuyin IME + Ctrl+Space switching - normal single input ✅
- [x] macOS Zhuyin IME + Fn switching - normal single input ✅

**Regression Testing:**
- [x] English keyboard input - normal operation ✅
- [x] Shift+Enter multiline input - works correctly ✅
- [x] Text paste (Cmd+Shift+V) - no duplicates ✅
- [x] Image paste - works correctly ✅
- [x] Basic command execution (ls, echo, etc.) - normal ✅
- [x] Cmd+K clear terminal - works correctly ✅
- [x] Copy selected text (Cmd+Shift+C) - works correctly ✅

## Demo

https://github.com/user-attachments/assets/8341cdf9-6c57-413e-b940-89e50cc79ff0

https://github.com/user-attachments/assets/d3a6e72a-f488-45c1-ab58-88391639455a

https://github.com/user-attachments/assets/ac178abd-caf3-40bf-9ef7-7cc0567a32c3

All features have been tested successfully on macOS with Claude Code in Wave Terminal.

### DIFF
--- a/pkg/wshrpc/wshserver/wshserver.go
+++ b/pkg/wshrpc/wshserver/wshserver.go
@@ -448,7 +448,14 @@ func (ws *WshServer) FileRestoreBackupCommand(ctx context.Context, data wshrpc.C
 func (ws *WshServer) GetTempDirCommand(ctx context.Context, data wshrpc.CommandGetTempDirData) (string, error) {
 	tempDir := os.TempDir()
 	if data.FileName != "" {
-		return filepath.Join(tempDir, data.FileName), nil
+		// Reduce to a simple file name to avoid absolute paths or traversal
+		name := filepath.Base(data.FileName)
+		// Normalize/trim any stray separators and whitespace
+		name = strings.Trim(name, `/\`+" ")
+		if name == "" || name == "." {
+			return tempDir, nil
+		}
+		return filepath.Join(tempDir, name), nil
 	}
 	return tempDir, nil
 }


sawka:
@togo01 thanks for working on this.  i think you're right that we should default the Shift+Enter case to work better with claude code rather than have it default to the less friendly version.  but i'm not sure about this change:

```
-                this.sendDataToController("\u001b\n");
+                this.sendDataToController("\n");
```

i think sending the ESC + "\n" is right (basically sends an Alt+Enter).  When I test with Claude Code, the ESC+\n inserts the newline correctly without submitting (also the same way that VSCode solved it with /term-setup).  Are you seeing something different?

togo01:
> @togo01 thanks for working on this. i think you're right that we should default the Shift+Enter case to work better with claude code rather than have it default to the less friendly version. but i'm not sure about this change:
> 
> ```
> -                this.sendDataToController("\u001b\n");
> +                this.sendDataToController("\n");
> ```
> 
> i think sending the ESC + "\n" is right (basically sends an Alt+Enter). When I test with Claude Code, the ESC+\n inserts the newline correctly without submitting (also the same way that VSCode solved it with /term-setup). Are you seeing something different?

Thanks for the feedback! You're absolutely right - I've reverted the change back to use \u001b\n (ESC + newline).
After reviewing the original implementation and testing again, I can confirm that \u001b\n is the correct approach. It properly inserts a newline in Claude Code without submitting the command, just like VSCode's /term-setup solution.
The feature works perfectly with this configuration. Thanks for catching that!

sawka:
overall this looks good, tested both the image pasting and the Shift+Enter code!  i'm going to merge it as-is and continue testing a bit more... i can fix anything that comes up in a follow-up PR rather than waiting forever to get it merged.

togo01:
> overall this looks good, tested both the image pasting and the Shift+Enter code! i'm going to merge it as-is and continue testing a bit more... i can fix anything that comes up in a follow-up PR rather than waiting forever to get it merged.

Thanks for the quick review and merge. I'm pleased to see this feature/fix incorporated.

PR #2677, opened 2025-12-15
Prevent AI panel auto-scroll when user scrolls up

stevenvo:
## Summary
Fixes the annoying behavior where the AI panel continuously auto-scrolls to the bottom while streaming, making it impossible to read earlier parts of long responses.

## Problem
When the AI is generating a long response, the panel auto-scrolls to show new content. If you try to scroll up to read something earlier, it immediately scrolls you back down, making it very frustrating to read.

## Solution
- Detects when user manually scrolls up (>50px from bottom)
- Stops auto-scrolling while user is reading earlier content
- Resumes auto-scrolling when user scrolls back near the bottom
- Uses passive event listener for better performance

## Test Plan
- [x] Start AI chat with long response
- [x] Scroll up while AI is streaming
- [x] Verify scroll position stays where you scrolled
- [x] Scroll back to bottom
- [x] Verify auto-scroll resumes


sawka:
this was a great idea (and definitely needed to be fixed).  i wrote my own fix for it here:

https://github.com/wavetermdev/waveterm/pull/2744

thank you for submitting it though

PR #2679, opened 2025-12-15
Add Cmd+Up/Down keyboard shortcuts for terminal scrolling

stevenvo:
## Summary
Adds keyboard shortcuts for quick navigation in terminal output.

## Shortcuts
- **Cmd+Up**: Scroll to top of terminal
- **Cmd+Down**: Scroll to bottom of terminal

## Use Cases
- Quickly jump to start of long command output
- Return to bottom after scrolling up to read
- Navigate large terminal buffers efficiently
- Useful when terminal has scrolled unexpectedly

## Implementation
- Added to terminal keyDownHandler
- Works in any terminal block
- Instant navigation (no animation)
- Doesn't interfere with other shortcuts

## Test Plan
- [x] Cmd+Up scrolls to top
- [x] Cmd+Down scrolls to bottom
- [x] Works with long terminal output
- [x] Doesn't break existing shortcuts

### DIFF
--- a/frontend/app/view/term/term-model.ts
+++ b/frontend/app/view/term/term-model.ts
@@ -489,17 +489,39 @@ export class TermViewModel implements ViewModel {
             this.setTermMode(newTermMode);
             return true;
         }
-        if (keyutil.checkKeyPressed(waveEvent, "Cmd:ArrowDown")) {
-            // Scroll to bottom
+        if (keyutil.checkKeyPressed(waveEvent, "Shift:End")) {
             if (this.termRef?.current?.terminal) {
                 this.termRef.current.terminal.scrollToBottom();
             }
             return true;
         }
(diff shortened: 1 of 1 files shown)


sawka:
@stevenvo I did some research here, and it isn't safe to bind Cmd+Up/Down (which ends up mapping to Alt+Up/Down) in the terminal.  This can conflict with terminal apps like `vim` or `less`.

So, I looked into what native MacOS and Linux/Windows terminals use.  And it turns out to be:

Shift:PageUp/Down, and Shift:Home/End which controls scrollback.  On MacOS you can get these by using Fn:Shift:Arrows (Fn:Shift:ArrowLeft is Shift:Home, Fn:Shift:ArrowUp is Shift:PageUp, etc).

These now mirror iterm2 and terminals like gnome terminal etc and so are safe to add.

PR #2709, opened 2025-12-22
some RPC client/server cleanup

sawka:
* announce and unannounce should propagate
* route the sendmessage command
* add timeouts for sending to OutputCh (appropriately)
* add ctx for SendCancel function

### DIFF
--- a/pkg/wshutil/wshrouter.go
+++ b/pkg/wshutil/wshrouter.go
@@ -355,9 +355,9 @@ func (router *WshRouter) UnregisterRoute(routeId string) {
 	router.Lock.Lock()
 	delete(router.RouteMap, routeId)
 	// clear out announced routes
-	for routeId, localRouteId := range router.AnnouncedRoutes {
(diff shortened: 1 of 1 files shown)


stevenvo:
Found a type mismatch bug introduced in this PR:

```
pkg/wshrpc/wshclient/wshclientutil.go:66:24: cannot use func(ctx context.Context) error {...} 
as func() value in assignment
```

**Issue**: Line 66 assigns a function with signature `func(ctx context.Context) error` to `opts.StreamCancelFn`, but `StreamCancelFn` is defined as `func()` in `wshrpctypes.go:376`.

**Current code** (line 66):
```go
opts.StreamCancelFn = func(ctx context.Context) error {
    return reqHandler.SendCancel(ctx)
}
```

**Should be**:
```go
opts.StreamCancelFn = func() {
    reqHandler.SendCancel(context.Background())
}
```

This causes build failures when building from latest main.

PR #2712, opened 2025-12-23
Improve search functionality

carl2027:
- Fixed: The search dialog now correctly receives focus when opened in the current block.
- Implemented direct search for selected text.


sawka:
ah, missed your PR with the holidays.  thanks for submitting, will take a look

PR #2714, opened 2025-12-25
Fix terminal state loss when switching workspaces

stevenvo:
## Summary
Fixes issue where TUI applications (vim, htop, opencode, etc.) would lose terminal state when switching between workspaces, causing inability to scroll and display corruption.

## Problem
When switching workspaces, all tab views were being destroyed and recreated. This caused:
- Loss of alternate screen buffer mode state
- Terminal display corruption with TUI apps
- Inability to scroll in terminals after returning to workspace
- Running TUI applications to appear broken

## Root Cause
The switchworkspace operation in emain/emain-window.ts called removeAllChildViews() which destroyed all tab views including their xterm.js terminal instances. When recreated from cache, the terminal state was not properly restored.

## Solution
Cache tab views across workspace switches instead of destroying them:
- Tab views are positioned off-screen but kept alive
- Terminal state fully preserved (buffer modes, scrollback, cursor position)
- Views only destroyed when tab explicitly closed or window closes
- Memory-efficient: ~1-5MB per cached terminal

## Changes
- Added allTabViewsCache map to preserve tab views during workspace switches
- Modified workspace switch logic to cache instead of destroy
- Added cache cleanup when tabs are explicitly closed
- Cache cleared when window closes to prevent memory leaks

## Testing
1. Run a TUI app (vim, htop, opencode) in workspace A
2. Switch to workspace B
3. Switch back to workspace A (even after hours)
4. Terminal state is preserved, scrolling works correctly

## Behavior Match
Now matches macOS Terminal, iTerm2, and other terminal emulators where terminals remain active across tab/workspace switches.

### DIFF
--- a/emain/emain-window.ts
+++ b/emain/emain-window.ts
@@ -139,6 +139,7 @@ export class WaveBrowserWindow extends BaseWindow {
     waveWindowId: string;
     workspaceId: string;
     allLoadedTabViews: Map<string, WaveTabView>;
+    allTabViewsCache: Map<string, WaveTabView>; // Cache for preserving tab views across workspace switches
     activeTabView: WaveTabView;
     private canClose: boolean;
     private deleteAllowed: boolean;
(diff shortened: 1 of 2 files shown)


stevenvo:
Closing temporarily for additional testing. Will reopen after validation.

PR #2715, opened 2025-12-25
Fix type mismatch in StreamCancelFn assignment

stevenvo:
## Summary
Fixes build error introduced in #2709 where `StreamCancelFn` assignment has incorrect function signature.

## Problem
Build fails with:
```
pkg/wshrpc/wshclient/wshclientutil.go:66:24: cannot use func(ctx context.Context) error 
as func() value in assignment
```

## Root Cause
- `StreamCancelFn` is defined as `func()` in `wshrpctypes.go:376`
- But was assigned a function with signature `func(ctx context.Context) error` in `wshclientutil.go:66`

## Fix
Match the expected signature by:
1. Removing parameters and return value from the lambda
2. Using `context.Background()` for the SendCancel call

## Testing
Build now succeeds without type errors.

## Related
- Introduced in: #2709
- Comment: https://github.com/wavetermdev/waveterm/pull/2709#issuecomment-3691587427

### DIFF
--- a/pkg/wshrpc/wshclient/wshclientutil.go
+++ b/pkg/wshrpc/wshclient/wshclientutil.go
@@ -63,8 +63,8 @@ func sendRpcRequestResponseStreamHelper[T any](w *wshutil.WshRpc, command string
 		rtnErr(respChan, err)
 		return respChan
 	}
-	opts.StreamCancelFn = func(ctx context.Context) error {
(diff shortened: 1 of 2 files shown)


stevenvo:
Closing to create new PR with accurate description. Initial analysis was incorrect - the fix is simpler than described.

PR #2725, opened 2025-12-28
add OSC 52 clipboard support for terminal applications

### LINKED ISSUE #2140: [Feature]: Support OSC 52 clipboard escape sequences over SSH"
### Feature description

 **Problem:** When connected to a remote machine using Wave Terminal's direct SSH connection, copying text from the remote terminal to the local system clipboard often fails, especially when terminal multiplexers like `tmux` or tools that use them internally (like `zsh4humans`) are running on the remote host. Standard mouse selection copy doesn't interact with the remote application's understanding of the clipboard in these cases.

**Reason:** Many modern terminal applications and tools (like `tmux` with `set-clipboard on`, `neovim`, etc.) utilize the OSC 52 escape sequence standard (`\e]52;...`) to programmatically interact with the terminal emulator's clipboard. This allows copying text even when mouse support is disabled or when copying across remote sessions. Wave Terminal does not currently appear to interpret or act on these OSC 52 sequences when received over a direct SSH connection.

### Implementation Suggestion

_No response_

### Anything else?

**Context:** This issue was identified when using `zsh4humans` over a direct SSH connection in Wave. `zsh4humans` optionally uses an integrated `tmux` session which enables `set-clipboard on` by default, relying on OSC 52 for clipboard operations. Disabling the integrated `tmux` (`zstyle ':z4h:' start-tmux no` in `~/.zshrc`) is a workaround but disables other `zsh4humans` features.

**Tmux Reference:** The `set-clipboard` option in `tmux` documentation explains its reliance on terminal clipboard support: [https://man.openbsd.org/tmux.1#set-clipboard](https://man.openbsd.org/tmux.1#set-clipboard)

**OSC 52 Explanation:** A good overview of how OSC 52 works can be found in various terminal emulator documentation or blog posts, such as this one: [https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands) (Search for "OSC 52") or [https://cirw.in/blog/bracketed-paste](https://cirw.in/blog/bracketed-paste) (mentions OSC 52 in context).

**Other Terminals:** Support for OSC 52 over SSH is present in terminals like iTerm2, Kitty, WezTerm, and others, making it a fairly standard feature for modern terminal emulators.

gaius-codius:
## Summary
  - Implements OSC 52 escape sequence handling for clipboard operations
  - Allows terminal apps like zellij, tmux, and neovim to copy to system clipboard
  - Fixes #2140

  ## Features
  - Write-only clipboard access (read queries blocked for security)
  - 75KB size limit matching common terminal implementations
  - Base64 whitespace handling per RFC 4648

  ## Additional Fix
  Also fixes StreamCancelFn type mismatch from #2709 where the implementations were updated but the type definition was not.

 - Tested with zellij mouse selection

### DIFF
--- a/frontend/app/view/term/term.tsx
+++ b/frontend/app/view/term/term.tsx
@@ -286,6 +286,7 @@ const TerminalView = ({ blockId, model }: ViewComponentProps<TermViewModel>) => 
                 keydownHandler: model.handleTerminalKeydown.bind(model),
                 useWebGl: !termSettings?.["term:disablewebgl"],
                 sendDataHandler: model.sendDataToController.bind(model),
+                nodeModel: model.nodeModel,
             }
         );
         (window as any).term = termWrap;
--- a/frontend/app/view/term/termwrap.ts
+++ b/frontend/app/view/term/termwrap.ts
@@ -1,20 +1,12 @@
 // Copyright 2025, Command Line Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+import type { BlockNodeModel } from "@/app/block/blocktypes";
 import { getFileSubject } from "@/app/store/wps";
 import { sendWSCommand } from "@/app/store/ws";
 import { RpcApi } from "@/app/store/wshclientapi";
 import { TabRpcClient } from "@/app/store/wshrpcutil";
-import {
-    WOS,
-    atoms,
-    fetchWaveFile,
-    getApi,
-    getSettingsKeyAtom,
-    globalStore,
-    openLink,
-    recordTEvent,
-} from "@/store/global";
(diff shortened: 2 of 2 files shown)


sawka:
@gaius-codius thanks for submitting this!  I've thought a bit about this clipboard support and I've struggled with what happens if this code gets emitted in a tab that isn't showing or just running some stuff in the background.

What do you think about not allowing the copy unless the block has focus?

gaius-codius:
Yes agree, I should have added a focus check originally. Thank you for adding this and for the great terminal application!

PR #2819, opened 2026-02-01
Update Model Mode Docs

ericmann:
The original `ai:model` name for xAI's models in the example didn't work as advertised (resulting a 404 error in practice). After some trial and error I've been able to sort out the correct model slug and have updated the documentation accordingly.


sawka:
thanks for submitting.  i think that bad model name must have been from openrouter.  definitely a bad example.  appreciate the fix.

PR #2821, opened 2026-02-03
Session Durability Checkpoint

sawka:
Working on bug fixes and UX.  Streams restarting, fixed lots of bugs, timing issues, concurrency bugs.  Get status shipped to the FE to drive "shield" state display.  Deal with stale streams.

Also big UX changes to the block headers.  Specialize the terminal headers to prioritize the connection (sense of place), remove old terminal icon and word "Terminal" from the header.  Also drop "Web" and "Preview" labels on web/preview blocks.

Added `wsh focusblock` command.


sawka:
@coderabbitai review

PR #2835, opened 2026-02-06
Tab background glow on process exit/bell

JustHereToHelp:
## Tab Background Glow + Indicator Improvements

Been using tab indicators with Claude Code across 8-10 tabs daily and kept running into friction — spinning icons are distracting, indicators go out of sync, and there's no way to tell at a glance which tabs are working vs done. With these fixes it is clear as day which tabs are running a command, which are done and need me to look at them, which are done and I have already looked at. Such a major help to my workflow.

This PR adds a `term:exitindicator` config option that lights up tab backgrounds based on process state, plus some quality-of-life improvements to how indicators feel:

**What it does:**
- Amber background glow while a cmd block is running
- Green glow on successful exit (exit 0), red on failure (non-zero/signal)
- `icon="none"` support — sets the background glow without showing an icon, for tools that just want a color state
- Synchronized breathing animation — all active tabs pulse together using a shared CSS `@property` variable on `:root` (16s cycle, barely perceptible)
- Static backgrounds for completed states (no animation on done/acknowledged tabs)
- Config: `term:exitindicator` (default false)

**The breathing sync was the tricky part.** Per-tab CSS animations drift out of phase. Solved it by animating a single `--breathe-phase` custom property on `:root` and having tabs read from it via `calc()` in the alpha channel. One clock, all tabs.

### Screenshot
Screenshot 2026-02-08 at 2 09 33 PM

Three states visible: grey check (done, already viewed), amber breathing (working), green check (done, needs attention).

**Files changed:**
- `pkg/blockcontroller/shellcontroller.go` — running + exit indicators for cmd blocks
- `frontend/app/tab/tab.tsx` — indicator-breathing class, icon="none" support
- `frontend/app/tab/tab.scss` — @property breathing, static vs animated states
- Config, schema, types, docs for `term:exitindicator`


JustHereToHelp:
Hey, thanks for the thorough review!

Just pushed a fix for both actionable items:

**CSS variable fallback** — good catch, I updated the condition to check `indicator != null` instead of `indicator?.color` so the CSS variable always gets set when the class is applied. Falls back to amber if color is somehow missing.

**Global config fallback** — this was actually already fixed in the latest push (the review might've run against the earlier commit). The exit goroutine checks block meta first with `HasKey`, then falls back to `wconfig.GetWatcher().GetFullConfig().Settings.TermExitIndicator`. Same pattern for the running indicator at process start.

**Hover glow** — fixed that too, added `.has-indicator` and `.has-indicator.active` overrides inside the hover block so the glow persists with the right opacity.

**Duplicated block-data fetch** — yeah I noticed that. Kept it separate intentionally since the exit indicator goroutine runs async and the close-on-exit check has its own lifecycle, but I agree it could be cleaner. Happy to refactor if the team prefers.

**Docs version reference** — that v0.11.5 header was already there before my change, didn't want to touch unrelated stuff in this PR. Can update it in a follow-up if you want.

JustHereToHelp:
### Updated: Tab indicator states in action

Three states visible — grey (done, already viewed), amber breathing (working), green (done, needs attention). Using this daily across 8-10 Claude Code tabs and it's been a massive workflow improvement.

PR #2858, opened 2026-02-11
fix: terminal scroll preservation on tab switch and alt buffer exit

### LINKED ISSUE #2384: [Bug]: Terminal scrolls to top after Claude output
### Current Behavior

When using Claude and arbitrary points, but always when Claude asks for input, the Wave terminal will scroll all the way to the top

### Expected Behavior

Don't scroll to the top :)

### Steps To Reproduce

1. Use wave
2. Run Claude
3. Ask claude to do something

### Wave Version

0.11.6

### Platform

macOS

### OS Version/Distribution

Tahoe 26.0

### Architecture

arm64

### Anything else?

_No response_

### Questionnaire

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

### LINKED ISSUE #2790: [Bug]: Terminal still scrolls to top after Claude output
### Current Behavior

When using Claude Code and when Claude asks for input (or sometimes randomly), the Wave Terminal will automatically scroll all the way to the top and user has to manually scroll down. I saw another issue describing the same already open since October 2nd. Unfortunately makes Wave Terminal almost unusable for Claude Code users.

### Expected Behavior

Doesn't scroll up.

### Steps To Reproduce

1. Use a Terminal inside Wave to use Claude Code

### Wave Version

0.13.1

### Platform

macOS

### OS Version/Distribution

Tahoe 26.2

### Architecture

arm64

### Anything else?

_No response_

### Questionnaire

- [x] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

JustHereToHelp:
Fixes #2384, fixes #2790

Switching tabs or quitting a fullscreen app (vim, htop, less) resets the scroll position to the top.

Tab switch: the intermediate repositioning during switch was triggering resize events that nuked the viewport. Now skips repositioning for initialized tabs and only restores scroll if dimensions actually changed.

Alt buffer exit: saves `baseY` before the buffer swap via `onBufferChange` and restores it after. Won't interfere if you were already at the bottom.

Mostly fixed — still see occasional jumps in edge cases.

### DIFF
--- a/emain/emain-window.ts
+++ b/emain/emain-window.ts
@@ -433,7 +433,7 @@ export class WaveBrowserWindow extends BaseWindow {
             await Promise.all([p1, p2]);
         } else {
             console.log("reusing an existing tab, calling wave-init", tabView.waveTabId);
-            const p1 = this.repositionTabsSlowly(35);
+            const p1 = this.repositionTabsSlowly(35, true);
             const p2 = tabView.webContents.send("wave-init", tabView.savedInitOpts); // reinit
             await Promise.all([p1, p2]);
         }
@@ -452,13 +452,16 @@ export class WaveBrowserWindow extends BaseWindow {
         }, 30);
     }
 
-    private async repositionTabsSlowly(delayMs: number) {
+    private async repositionTabsSlowly(delayMs: number, skipIntermediate: boolean = false) {
         const activeTabView = this.activeTabView;
         const winBounds = this.getContentBounds();
         if (activeTabView == null) {
             return;
         }
-        if (activeTabView.isOnScreen()) {
+        if (skipIntermediate || activeTabView.isOnScreen()) {
+            // For already-initialized tabs (skipIntermediate=true) or tabs already on screen,
+            // go directly to final bounds to avoid an intermediate resize that causes
+            // xterm.js buffer reflow and scroll position loss.
             activeTabView.setBounds({
                 x: 0,
                 y: 0,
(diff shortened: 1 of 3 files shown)


sawka:
working on the bug, but different fixes.  #2944 and #2941 .  @JustHereToHelp are you on Linux or Windows?  I can't seem to reproduce this on Mac.  My guess is that it is related to showing/hiding the scrollbar which might cause a reflow (but even on Windows I can't repro it).  I also have never seen a resize event fire after the "repositionTabsSlowly" or when switching tabs.  So I'm really confused what is going on.  Is the "repositionTabsSlowly" fix required to fix your issue?

danbuhler:
The problem doesn't happen to me all the time. I'm on 0.14.0 202602112358 on MacOS Tahoe 26.3. But https://github.com/wavetermdev/waveterm/pull/2941 does happen to me all the time. I wonder if they're related. I'll wait to see if that fix solves all my issues.

JustHereToHelp:
Hey! Sorry for the late reply. I'm on macOS — it was happening in my dev build. I saw v0.14.1 shipped the Claude Code scroll fix and your PRs #2941 and #2944 addressed it from the upstream side too. Pretty sure the repositionTabsSlowly fix in this PR isn't needed anymore given all that work.

Happy to close this if you think it's fully covered now — just wanted to check with you first since @danbuhler also reported seeing it intermittently on Mac. Either way, really happy with how the scroll behavior feels in the latest releases and just happy in general, loving the shipments!!

danbuhler:
It's fixed! Loving it! @JustHereToHelp

PR #2859, opened 2026-02-11
feat: cmd+click to open file paths in terminal

JustHereToHelp:
## Summary

Added cmd+click detection for file paths in the terminal. Click any file path and it opens in Wave's preview block. Way more useful than I expected — especially when Claude or build tools spit out file paths in error messages.

Implementation:
- Custom `FilePathLinkProvider` registered via xterm.js `registerLinkProvider` API
- Matches absolute paths, relative paths, `~/` paths, and `file:line:col` patterns
- Opens files in Wave's preview block (not an external editor) using the existing `preview:` block type
- Registered in `termwrap.ts`, provider logic in new file `term-link-provider.ts`

Patterns matched:
- `/absolute/path/to/file.ts`
- `./relative/path`
- `~/home/path`
- `/path/to/file.ts:42:10` (file:line:col)

Builds on the hyperlink support from #1357.

## Test plan
- Cmd+click an absolute file path in terminal output — should open in preview
- Cmd+click a `~/` path — should resolve and open
- Cmd+click a `file:line:col` pattern (e.g., from TypeScript errors) — should open
- Verify regular text selection still works (no false path matches on normal text)
- Verify link underline appears on hover with Cmd held

### DIFF
--- a/frontend/app/view/term/term-link-provider.ts
+++ b/frontend/app/view/term/term-link-provider.ts
@@ -18,7 +18,10 @@ const FILE_PATH_REGEX =
 
 // File extensions we recognize for bare relative paths (the ones without ./ prefix)
 const KNOWN_EXTENSIONS =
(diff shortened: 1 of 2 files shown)


JustHereToHelp:
Thanks for the catch on both of these.

Fixed in 1bb0806:
- **IDisposable leak**: `registerLinkProvider` return value is now tracked in `toDispose` so it gets cleaned up with everything else on terminal disposal.
- **Makefile/Dockerfile**: Moved these out of `KNOWN_EXTENSIONS` (which requires a leading dot that they'll never have) into a separate `KNOWN_FILENAMES` regex that matches by exact basename. Added a few other common extensionless files too — Rakefile, Gemfile, Justfile, etc.

PR #2902, opened 2026-02-20
fix: pass ProxyURL from AI mode config to backend

sg-shag:
## Problem

`ai:proxyurl` set in AI mode configurations (Wave AI panel) has no effect.

## Root Cause

`ProxyURL` field is missing from `AIModeConfigType` (the new Wave AI mode config struct), and is not passed through when converting to `AIOptsType` in `buildAIOptsFromMode()`.

The proxy support code already exists in both OpenAI (`openai-backend.go:535`) and Anthropic (`anthropic-backend.go:461`) backends — they read `chatOpts.Config.ProxyURL` — but the value was never populated.

## Fix

Two lines:
1. Add `ProxyURL` field to `AIModeConfigType` in `settingsconfig.go`
2. Pass `config.ProxyURL` to `AIOptsType` in `usechat.go`

## Testing

With this fix, adding `"ai:proxyurl": "http://127.0.0.1:8888"` to an AI mode config correctly routes requests through the specified HTTP proxy.

### DIFF
--- a/pkg/aiusechat/usechat.go
+++ b/pkg/aiusechat/usechat.go
@@ -123,6 +123,7 @@ func getWaveAISettings(premium bool, builderMode bool, rtInfo waveobj.ObjRTInfo,
 		Verbosity:     verbosity,
 		AIMode:        aiMode,
 		Endpoint:      baseUrl,
+		ProxyURL:      config.ProxyURL,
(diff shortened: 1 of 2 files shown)


sawka:
handling this in #2961 ... the proxy url had to be passed through to multiple providers.  thanks for the PR/bug report!

Issue #2915, opened 2026-02-21
Korean IME composition not committed — characters only appear after pressing Space


juliuschun:
## Description

Korean characters are not committed during normal typing. 
The composition remains in a "pending" state indefinitely, and characters 
only appear in the terminal when pressing Space (which forces compositionend).

Typing "안녕하세요" results in nothing appearing until Space is pressed, 
at which point some characters are lost/skipped.

## Environment

- **Wave Terminal version:** latest (autoupdate:latest channel)
- **macOS version:** Darwin 25.2.0 (Sequoia)
- **Input method:** macOS built-in Korean (두벌식)
- **Locale:** ko_KR.UTF-8
- **Term settings:** rendermode: dom, font: JetBrains Mono

## Steps to Reproduce

1. Open Wave Terminal
2. Switch to Korean input (macOS built-in 한글)
3. Type any Korean text (e.g., "테스트")
4. Observe: characters stay in composition state, not committed
5. Press Space → characters finally appear (with some dropped)

## Expected Behavior

Korean characters should be committed to the terminal as each syllable 
is completed (same as VS Code integrated terminal, iTerm2, etc.)

## Comparison

- **VS Code terminal (also Electron + xterm.js):** Works correctly ✅
- **iTerm2:** Works correctly ✅
- **Wave Terminal:** Broken ❌

This suggests the issue is in Wave's input event handling layer 
between Electron and xterm.js, not in xterm.js itself.

## Possibly Related

- v0.12.3 fixed "duplicate text issue when switching input methods during CJK composition" — but the underlying composition commit issue persists
- Electron IME bugs: electron/electron#32307, electron/electron#29875
- Similar issue in Warp Terminal was fixed: warpdotdev/Warp#6891

sawka:
thanks for this bug report.  spent my day researching and testing IME.  i can easily reproduce your issue.  i have a fix in #2938 .  the trick was allowing korean input AND also fixing this weird input duplication bug with CapsLock from #2523 . should be part of the next patch release

sawka:
fixed in v0.14.1

astraltear:
### Describe the Bug
When typing in Korean (IME), the last character of a word is not properly committed to the terminal buffer. 
If I press the **Enter** key immediately after typing a word, the last character is lost (dropped). 
To prevent this character loss, I have to press the **Spacebar** to commit the composition before pressing **Enter**.

### Steps to Reproduce
1. Open Wave Terminal.
[...]

PR #2938, opened 2026-02-25
Fix IME issues

sawka:


sawka:
addresses #2915 and #2523

PR #2956, opened 2026-02-27
Fix for Claude Code Scroll to Top Bug

sawka:
There may be more cases here that I don't know about, but this fixes a good chunk of them.  This catches the CC "repaint" transaction and forces a scrollToBottom.  That should handle context repaints and resize repaints.

Also adds a new (hidden) terminal escape sequence debugger, and (in dev mode) adds a last 50 writes cache that can be used to look at and debug output.


sawka:
addresses #2790 and #2384

PR #3000, opened 2026-03-06
fix: notify Electron when active tab changes via SetActiveTab

Pwacca:
## Summary
- `SetActiveTab` service method was missing the `SendActiveTabUpdate` call to notify Electron about the active tab change
- This caused the frontend to not reflect the tab switch when `SetActiveTab` was called programmatically (e.g., via the API)
- Adds the missing `wcore.SendActiveTabUpdate(ctx, workspaceId, tabId)` call, consistent with how other tab-switching paths work

## Test plan
- [ ] Call `SetActiveTab` via the API and verify the UI switches to the correct tab
- [ ] Verify normal tab clicking still works as expected
- [ ] Verify no duplicate notifications are sent

### DIFF
--- a/pkg/service/workspaceservice/workspaceservice.go
+++ b/pkg/service/workspaceservice/workspaceservice.go
@@ -197,6 +197,7 @@ func (svc *WorkspaceService) SetActiveTab(workspaceId string, tabId string) (wav
 	if err != nil {
 		return nil, fmt.Errorf("error setting active tab: %w", err)
 	}
+	wcore.SendActiveTabUpdate(ctx, workspaceId, tabId)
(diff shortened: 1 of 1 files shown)


sawka:
@Pwacca This is not correct.  The correct flow is calling SendActiveTabUpdate which then calls into electron, which then calls back into workspaceservice.  Also there are no backend callers of WorkspaceService. SetActiveTab so not sure what this is trying to fix.

Are you trying to use the REST API independently?  That's not supported (the services API is quite old).  The modern way is to use the RPC system (you can see examples of how this is done in the `wsh` commands).

PR #3099, opened 2026-03-21
fix: warn before discarding unsaved changes in waveconfig view

### LINKED ISSUE #2890: [Bug]:  BYOK menu loses state when clicking away without warning - When I click away (into another sibling menu option) from the Wave AI Models tab without saving it it loses the state
### Current Behavior

When I click away (into another sibling menu option) from the Wave AI Models tab without saving it it loses the state

### Expected Behavior

Keep state, so I can go back to it or warn me that I'll lose my unsaved data from that field

### Steps To Reproduce

Self-evident

### Wave Version

Client Version 0.14.0 (202602120004) Update Channel: latest

### Platform

Windows

### OS Version/Distribution

Windows 10 22H2 build 19045.4412

### Architecture

x64

### Anything else?

Thank you. Great work, guys! No other bugs so far.

majiayu000:
Fixes #2890

## Changes

Add unsaved changes confirmation when navigating away in the Wave Config view. Two navigation paths lose edits without warning:

1. **File sidebar selection**: `handleFileSelect` calls `model.loadFile(file)` directly, which resets `hasEditedAtom` and overwrites `fileContentAtom` — discarding edits silently.

2. **JSON→Visual tab switching**: `setActiveTab()` is called directly. The visual component renders from saved config (not the editor buffer), so edits appear lost.

### Fix

- Added `confirmDiscardChanges()` to `WaveConfigViewModel` — checks `hasEditedAtom`, shows `window.confirm()` if there are unsaved changes.
- Guarded `handleFileSelect` with the confirmation check (skips if already on the selected file).
- Guarded JSON→Visual tab switch with the confirmation check.

Only prompts when `hasEditedAtom` is true. Visual→JSON direction is safe (editor loads from `fileContentAtom`). `window.confirm()` is synchronous and works well in Electron.

## Test Plan

- Edit JSON, switch file in sidebar → confirmation dialog appears. Cancel stays, confirm switches.
- Edit JSON, click Visual tab → confirmation dialog appears.
- Save changes first, then switch → no dialog.
- Switch without editing → no dialog.
- `npx tsc --noEmit` passes.


sawka:
thanks (again) for the PR!  looks good and fixes the bug

PR #3121, opened 2026-03-25
fix: set XDG_CONFIG_DIRS and XDG_DATA_DIRS defaults in tryGetPamEnvVars

### LINKED ISSUE #2970: [Bug]: Not Loading the XDG_CONFIG_DIRS Environment Variable
### Current Behavior

When I run "echo $XDG_CONFIG_DIRS" nothing is returned.

### Expected Behavior

Running echo $XDG_CONFIG_DIRS should print the results of loading /etc/X11/Xsession.d/60x11-common_xdg_path.  In my case, as I am running the KDE desktop, that would include a mention of "plasma" somewhere.

### Steps To Reproduce

1. Running the latest Kubuntu LTS version, run "echo $XDG_CONFIG_DIRS" from the command line.
2.  It should result in something like: /home/myuser/.config/kdedefaults:/etc/xdg/xdg-plasma:/etc/xdg:/usr/share/kubuntu-default-settings/kf5-settings
3. The reason I noticed this is the results of running fastfetch or neofetch shows the OS as being Ubuntu instead of Kubuntu.

### Wave Version

v0.14.0

### Platform

Linux

### OS Version/Distribution

Kubuntu 24.04.4

### Architecture

x64

### Anything else?

_No response_

majiayu000:
## Summary

- Fixes #2970: WaveTerm does not inherit `XDG_CONFIG_DIRS` (and `XDG_DATA_DIRS`) when snap or other environments strip these variables and PAM env files do not define them
- In `tryGetPamEnvVars()`, after the existing `XDG_RUNTIME_DIR` fallback, adds identical fallback logic for `XDG_CONFIG_DIRS` (default: `/etc/xdg`) and `XDG_DATA_DIRS` (default: `/usr/local/share:/usr/share`) per the XDG Base Directory Specification
- No behavior change when these vars are already set by PAM env files

## Root Cause

Snap confinement strips several XDG environment variables. `tryGetPamEnvVars()` already handles `XDG_RUNTIME_DIR` with a sensible default, but `XDG_CONFIG_DIRS` and `XDG_DATA_DIRS` were left unhandled, causing child shells to receive empty/unset values.

## Test plan

- [ ] `gofmt -l ./pkg/shellexec/` — no output (clean)
- [ ] `go build ./...` — succeeds
- [ ] On Linux with snap, verify that child shells receive `XDG_CONFIG_DIRS=/etc/xdg` and `XDG_DATA_DIRS=/usr/local/share:/usr/share` when the variables are not set by the desktop environment

### DIFF
--- a/pkg/shellexec/shellexec.go
+++ b/pkg/shellexec/shellexec.go
@@ -760,5 +760,11 @@ func tryGetPamEnvVars() map[string]string {
 	if runtime_dir, ok := envVars["XDG_RUNTIME_DIR"]; !ok || runtime_dir == "" {
 		envVars["XDG_RUNTIME_DIR"] = "/run/user/" + fmt.Sprint(os.Getuid())
 	}
+	if configDirs, ok := envVars["XDG_CONFIG_DIRS"]; !ok || configDirs == "" {
+		envVars["XDG_CONFIG_DIRS"] = "/etc/xdg"
+	}
+	if dataDirs, ok := envVars["XDG_DATA_DIRS"]; !ok || dataDirs == "" {
+		envVars["XDG_DATA_DIRS"] = "/usr/local/share:/usr/share"
+	}
 	return envVars
 }


sawka:
@majiayu000 thanks again for the submission.  looks good!

PR #3134, opened 2026-03-27
Add custom keybindings + widgetBar toggle + few fixes

mswiszcz:
First of all - great project! I think this has potential to be the best terminal emulator on the market.

This PR includes few adjustments I wanted to add:
- Option to toggle sidebar in the workspace - I don't use it that often so I don't want it to waste the space. It's remembering the state across sessions and workspaces
- Option to set custom keybindings, following VSCode patterns
- Support for Mouse-3 and Mouse-4 buttons in browser
- 
And few fixes:
- Fix for browser stealing keyboard shortcuts - when browser was focused, switching between waveterm tabs with CMD+Number was not possible
- Fix for washed out colors when using some terminal tools (missing COLORTERM env variable)
- Fix for empty space if AI button is disabled, which is sometimes rendered and sometimes not. See bug on screenshot:
Screenshot 2026-03-27 at 17 37 24

There are few adjustments to the keybindings I want to make here, so I am for now keeping this as a draft.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "productName": "Wave",
     "description": "Open-Source AI-Native Terminal Built for Seamless Workflows",
     "license": "Apache-2.0",
-    "version": "0.14.4-beta.2",
+    "version": "0.14.4",
     "homepage": "https://waveterm.dev",
     "build": {
         "appId": "dev.commandline.waveterm"


sawka:
@mswiszcz thanks for your interest in the project, but there is just no way i can merge a PR like this.  please read the contributing guidelines.

for something like keybindings, which is a major feature, and creates an ongoing user-facing dependency, we'd need to have a discussion in discord at the very least.

for the other features you've added, they need to be self contained in separate PRs so they can be tested and merged independently. i'd definitely be open to merging the webview mouse changes, and the widgets sidebar hiding change if they came in as separate PRs.

mswiszcz:
@sawka sure I'll extract the keybindings, sibear toggle and leave this pr with only the bugfixes - I'll update the pr later and send 2 others too :)

PR #3154, opened 2026-03-30
fix: allow non-ASCII characters in SSH host aliases

### LINKED ISSUE #2937: [Bug]: SSH connection and connections.json mapping fail when Host alias contains non-ASCII (Chinese) character
### Current Behavior

When defining an SSH connection alias (Host) in ~/.ssh/config that includes UTF-8 characters (e.g., Chinese characters like PROD-服务器), Wave Terminal fails to properly parse the alias or map it to the corresponding configurations in ~/.waveterm/config/connections.json. This causes automatic password injection (ssh:passwordsecretname), wsh initialization, and theme applications to silently fail or the connection to be completely unreachable.

### Expected Behavior

Wave Terminal should correctly parse UTF-8 / non-ASCII characters in the ~/.ssh/config file and accurately map them to the corresponding keys in connections.json, applying the theme, password, and wsh settings as it does for pure English aliases.

Workaround / Clue for maintainers
If I strictly rename the Host to use purely English characters (e.g., changing PROD-Baoguan-阿里云 to PROD-Baoguan-Aliyun) in both ~/.ssh/config and connections.json, everything works flawlessly. The password injects successfully, and the theme applies perfectly.
This heavily implies that the regex or the parsing library Wave uses to read the SSH config or map the JSON keys is failing to handle UTF-8 characters properly.

### Steps To Reproduce

Add an entry in ~/.ssh/config with a Chinese suffix in the Host name.

Map this exact Host name in ~/.waveterm/config/connections.json to apply a theme and a secret password.

Open Wave Terminal, click the connection dropdown, and select PROD-Baoguan-阿里云.

The connection fails to establish, or it prompts for a password, completely ignoring the ssh:passwordsecretname and term:theme defined in connections.json.

### Wave Version

Client Version 0.14.0 (202602112358) Update Channel: latest

### Platform

macOS

### OS Version/Distribution

_No response_

### Architecture

arm64

### Anything else?

_No response_

majiayu000:
Fixes #2937

`userHostRe` in `pkg/remote/connutil.go` only matched `[a-zA-Z0-9]` for the host portion, so SSH Host aliases with Chinese (or other non-ASCII) characters failed validation.

Added `\p{L}` and `\p{N}` Unicode classes to the regexp so it accepts Unicode letters and digits in both host and user parts. Go's regexp package supports these natively.

Tests added in `pkg/remote/connutil_test.go` covering Chinese, Japanese, mixed, and ASCII-only inputs.

### DIFF
--- a/pkg/remote/connutil_test.go
+++ b/pkg/remote/connutil_test.go
@@ -20,6 +20,7 @@ func TestParseOpts(t *testing.T) {
 		{"host only", "myserver", "", "myserver", "", false},
 		{"chinese host alias", "PROD-服务器", "", "PROD-服务器", "", false},
 		{"mixed ascii and chinese with user and port", "user@PROD-阿里云:22", "user", "PROD-阿里云", "22", false},
+		{"unicode user and host", "用户@服务器:22", "用户", "服务器", "22", false},
 		{"unicode only host", "服务器", "", "服务器", "", false},
 		{"japanese host", "サーバー", "", "サーバー", "", false},
 		{"empty string", "", "", "", "", true},


sawka:
@majiayu000 for this one, i think it is still not allowed to have non-ascii chars in the username.  technically host names can't have non-ascii either, but since we look up the host in ssh_config, ssh config might re-map the host using a HostName directive to ascii.

so, we want only ascii for user, and then allow unicode for host.

PR #3156, opened 2026-03-31
feat: Implement i18n internationalization support with dynamic transl…

re2zero:
…ation loading

- Install i18next, react-i18next, i18next-http-backend, i18next-browser-languagedetector
- Support English (en) and Chinese (zh-CN) languages
- Implement dynamic translation file scanning using import.meta.glob
  - Auto-discover all translation JSON files in translations/
  - Automatically register namespaces from file structure
  - Zero-config: adding new translations requires no code changes
- Create translation JSON files for 5 namespaces:
  - common: Common UI strings
  - waveconfig: Wave config interface
  - widgets: Widget related strings
  - settings: Settings page
  - filemenu: File menu strings
- Implement useT() hook with automatic namespace resolution
- Implement useAppLanguage() hook for language management
- Add language switching functionality without closing config
- Apply i18n translations to multiple components:
  - waveconfig: Settings content with language selection
  - widgets: Settings and help strings
  - preview-directory: File menu translations
- Configure publicDir in electron.vite.config.ts for static assets


sawka:
Thanks for the PR, I appreciate the work here, but I don’t want to take on full i18n right now. The ongoing maintenance cost is pretty high, since every new or changed UI string becomes part of the localization surface area. At this stage I’d rather keep the UI moving quickly than add that tax everywhere.

Also, if we do language support later, I’d prefer it to be explicitly user-selected rather than auto-detected.

PR #3158, opened 2026-03-31
new keybindng F2 to edit the active tab name, also fix refocus after name editing

sawka:


hovo1990:
Sorry but this is not a good idea at all, it messes up byobu, cannot create new tab in byobu.
And it does not look it is possible to change.

PR #3278, opened 2026-05-04
fix(UI)): ensure terminal theme application via xterm theme setter

jhaumont:
When you try to set a theme on a widget:
```json
"test": {
    "icon": "dharmachakra",
    "label": "test",
    "color": "#326ce5",
    "blockdef": {
      "meta": {
        "view": "term",
        "controller": "cmd",
        "cmd": "wsl.exe -d Debian --shell-type login bash -ic \"k9s\"",
        "cmd:interactive": true,
        "cmd:runonstart": true,
        "term:theme": "Dracula"
      }
    }
``` 

It's not take in account. You have to manually change the theme of the block.

Purpose of this PR is to fix this and having the widget taking into account the theme from start.


jhaumont:
@coderabbitai review

PR #3331, opened 2026-05-24
[codex] Fix terminal IME composition handling

dandacompany:
## Summary

- Preserve IME composition state when adapting keyboard events.
- Avoid Wave global/custom terminal key handling while IME composition is active.
- Add broader CJK font fallback while keeping Hack as the primary terminal font.
- Fix the `.fixed-font` CSS declaration to use a valid `font:` property.

## Why

IME composition key events can flow through Wave keyboard handling and terminal custom key handling. In interactive terminal prompts, this can cause unexpected input while composing text with Korean, Japanese, Chinese, or other composition-based input methods.

The terminal font fallback also previously relied only on Hack/monospace. Keeping Hack first preserves the existing monospace behavior, while adding Noto CJK and common platform CJK fallbacks improves glyph coverage for multilingual terminal output.

## Verification

- `npm run build:dev` in a clean worktree from this commit
- Build completed with exit code 0
- Existing Vite warnings observed: browser externalization for `fs`/`path`, circular chunk warning for cytoscape/mermaid
- Image optimizer reported missing optional `sharp` package, but the build command still completed successfully

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -199,6 +199,7 @@ tasks:
             - task: build:server:internal
               vars:
                   ARCHS: arm64,amd64
+                  GO_ENV_VARS: MACOSX_DEPLOYMENT_TARGET=12.0
 
     build:server:quickdev:
         desc: Build the wavesrv component for quickdev (arm64 macOS only, no generate).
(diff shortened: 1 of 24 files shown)


dandacompany:
@coderabbitai review

PR #3402, opened 2026-06-29
fix(shellexec): avoid false Snap detection and empty XDG_* vars

### LINKED ISSUE #3336: [Bug]: shellexec writes empty XDG_*_HOME when $SNAP is inherited from any snap (not just Wave-as-snap)
## Description

`wavesrv` writes **empty strings** for `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, and
`XDG_CACHE_HOME` into the environment of every shell it spawns, even when none
of these variables were set in `wavesrv`'s own environment. Tools that path-join
these vars without spec-compliant fallback (e.g. `mise`) then write their cache,
config, and data into the user's **current working directory** as relative
dirs (`./mise/`, `./claude/versions/`, `./applications/…`).

Originally I assumed this was a Wave-snap-only issue (similar to #3316), but
it reproduces on the **dev build** running from `electron-vite preview` — see
root cause below.

## Root cause

`pkg/shellexec/shellexec.go:646-666`:

```go
/*
  For Snap installations, we need to correct the XDG environment variables as Snap
  overrides them to point to snap directories. We will get the correct values, if
  set, from the PAM environment. If the XDG variables are set in profile or in an
  RC file, it will be overridden when the shell initializes.
*/
if os.Getenv("SNAP") != "" {
    log.Printf("Detected Snap installation, correcting XDG environment variables")
    varsToReplace := map[string]string{
        "XDG_CONFIG_HOME": "", "XDG_DATA_HOME": "", "XDG_CACHE_HOME": "",
        "XDG_RUNTIME_DIR": "", "XDG_CONFIG_DIRS": "", "XDG_DATA_DIRS": "",
    }
    pamEnvs := tryGetPamEnvVars()
    if len(pamEnvs) > 0 {
        for k := range pamEnvs {
            if _, ok := varsToReplace[k]; ok {
                varsToReplace[k] = pamEnvs[k]
            }
        }
    }
    log.Printf("Setting XDG environment variables to: %v", varsToReplace)
    shellutil.UpdateCmdEnv(ecmd, varsToReplace)
}
```

Two compounding bugs:

### Bug 1: Snap detection is too coarse

`os.Getenv("SNAP") != ""` returns true whenever `wavesrv` is launched from *any*
snap process, not just when Wave itself is installed as a snap. Lots of CLI
tools ship as snaps (mine: `task`, see env dump below), and snap classic
confinement propagates `$SNAP`, `$SNAP_NAME`, `$SNAP_DATA`, etc. all the way
down the process tree.

A more accurate detection would be:

```go
if os.Getenv("SNAP_NAME") == "waveterm" {
```

### Bug 2: Empty-string defaults are written verbatim when PAM env lacks XDG vars

`tryGetPamEnvVars()` parses `/etc/environment`, `/etc/security/pam_env.conf`,
and `~/.pam_environment`. On most modern systems (especially Ubuntu / Debian)
**none** of these set the XDG base-directory vars — XDG defaults are handled at
the freedesktop spec level, not via PAM. So `pamEnvs` either returns an empty
map or a map that doesn't contain any of the six XDG keys.

When that happens, the `varsToReplace` map keeps its empty-string defaults, and
`shellutil.UpdateCmdEnv(ecmd, varsToReplace)` writes them verbatim into the
spawned shell's environment.

Tools then path-join these empty values and write to relative paths in CWD.

## Reproduction (dev build)

Wave dev build, launched as `electron-vite preview` from a shell that has `$SNAP`
set due to a launching snap tool (here: a snap named `task`).

```bash
# Process tree
$ pstree -ps $(pgrep wavesrv.x64)
... electron-vite preview ─ electron ─ wavesrv.x64 ─ bash

# wavesrv's own environ — $SNAP is set (inherited from launching snap), but
# the three XDG_*_HOME vars themselves are NOT present
$ tr '\0' '\n' < /proc/$(pgrep wavesrv.x64)/environ | grep -E '^(SNAP=|XDG_)' | sort
SNAP=/snap/task/283
XDG_CONFIG_DIRS=/etc/xdg/xdg-ubuntu:/etc/xdg
XDG_CURRENT_DESKTOP=ubuntu:GNOME
XDG_DATA_DIRS=/usr/share/ubuntu:/usr/share/gnome:/home/USER/.local/share/flatpak/...
XDG_MENU_PREFIX=gnome-
XDG_RUNTIME_DIR=/run/user/1000
XDG_SESSION_CLASS=user
XDG_SESSION_DESKTOP=ubuntu
XDG_SESSION_TYPE=wayland

# bash spawned by wavesrv — the three *_HOME vars are now set-but-empty,
# *_DIRS got truncated, RUNTIME_DIR preserved
$ tr '\0' '\n' < /proc/<spawned-bash-pid>/environ | grep '^XDG_' | sort
XDG_CACHE_HOME=
XDG_CONFIG_DIRS=/etc/xdg
XDG_CONFIG_HOME=
XDG_CURRENT_DESKTOP=ubuntu:GNOME
XDG_DATA_DIRS=/usr/local/share:/usr/share
XDG_DATA_HOME=
XDG_MENU_PREFIX=gnome-
XDG_RUNTIME_DIR=/run/user/1000
XDG_SESSION_CLASS=user
XDG_SESSION_DESKTOP=ubuntu
XDG_SESSION_TYPE=wayland
```

From within any Wave shell:

```bash
$ printf 'XDG_CACHE_HOME=[%s] set?=[%s]\n' "$XDG_CACHE_HOME" "${XDG_CACHE_HOME-UNSET}"
XDG_CACHE_HOME=[] set?=[]
```

If unset, `set?` would print `UNSET`; it prints empty, proving the var is
explicitly set to an empty string.

## Impact (real-world)

I traced this after `mise outdated` started writing its cache into project
directories. Auditing `~/Projects/` turned up **686 MB** of leaked tool data
across 7 project repos:

| Tool                      | Leaked dir                                              | Cause                                              |
|---------------------------|---------------------------------------------------------|----------------------------------------------------|
| `mise outdated`           | `./mise/` (× 7)                                         | `mise` path-joins `XDG_CACHE_HOME + "mise"`        |
| Claude Code (autoupdater) | `./claude/versions/` (× 3, ~229 MB each)                | path-joins `XDG_DATA_HOME + "claude"`              |
| Claude Code (URL handler) | `./applications/claude-code-url-handler.desktop` (× 6)  | writes desktop entry to `$XDG_DATA_HOME/applications` |
| Android SDK Manager       | `./.android/cache/`                                     | metadata XMLs duplicated into CWD                  |

Every `cd <project>` followed by *any* of these tools running silently created
a fresh cache copy in the project. Three of them were untracked-but-not-
gitignored and would have been committed by an unwary `git add .`.

## Suggested fix

Two independent patches that should both land:

1. **Tighten snap detection** — gate on the Wave snap specifically, not any
   ancestor snap:

   ```go
   if os.Getenv("SNAP_NAME") == "waveterm" {
   ```

2. **Don't write empty-string defaults** — when `pamEnvs` lacks a given XDG
   key, either leave the var alone (so the spec default applies) or set it to
   the spec default explicitly:

   ```go
   xdgDefaults := map[string]string{
       "XDG_CONFIG_HOME": filepath.Join(home, ".config"),
       "XDG_DATA_HOME":   filepath.Join(home, ".local/share"),
       "XDG_CACHE_HOME":  filepath.Join(home, ".cache"),
   }
   for k, def := range xdgDefaults {
       if v, ok := pamEnvs[k]; ok && v != "" {
           varsToReplace[k] = v
       } else {
           varsToReplace[k] = def
       }
   }
   ```

   (Or just `delete(varsToReplace, k)` for keys with no PAM value, so the var
   stays unset and the child shell falls back via the spec.)

## Workaround

Add to `~/.bashrc` (after the interactive-shell guard):

```bash
[ -z "${XDG_CACHE_HOME:-}" ] && export XDG_CACHE_HOME="$HOME/.cache"
[ -z "${XDG_CONFIG_HOME:-}" ] && export XDG_CONFIG_HOME="$HOME/.config"
[ -z "${XDG_DATA_HOME:-}" ]  && export XDG_DATA_HOME="$HOME/.local/share"
```

This protects interactive shells. Non-interactive subprocesses launched from a
Wave shell (cron, systemd user units, CLI tools that fork before sourcing rc
files) still need per-script guarding.

## Environment

- WaveTerm: dev build, `electron-vite preview` from main (commit current as of
  2026-05-27). Binary at `~/Projects/waveterm/dist/bin/wavesrv.x64`.
- The wavesrv process had inherited `SNAP=/snap/task/283` and friends from a
  CLI snap (`task`) that was active in the launching shell, despite Wave itself
  not being installed as a snap.
- Distro: Ubuntu 26.04, Wayland / GNOME
- Shell: bash 5.x (system)

## Related

- #3316 (open) — `desktop-common.sh` deletes XDG user dirs / self-referential
  symlinks under classic confinement (a different XDG mishandling path)
- #2970 (closed) — XDG_CONFIG_DIRS not loaded
- #1696 (closed) — Snap isolation rewrites XDG directory paths

Jason-Shen2:
## Summary

This fixes #3336 with two tightly-scoped changes in `pkg/shellexec/shellexec.go`:

### 1. Narrow the Snap detection guard
The previous check `os.Getenv("SNAP") != ""` fires for **any** process that inherits $SNAP from a parent — including when Wave is launched from inside a Snap-installed terminal/CLI (e.g. the user snaps Alacritty or nvim and launches Wave from there). In that case Wave itself is not running as a Snap, but the code still wipes the XDG variables and tries to repopulate them from PAM.

Changed the guard to:

```go
if os.Getenv("SNAP_NAME") == "waveterm" {
```

`SNAP_NAME` is set by Snapd to the declared snap name for the **running** snap only; inherited $SNAP from a parent process does not carry the parent's `SNAP_NAME` across exec when the child is not a snap. This matches the documented Snap environment-variable conventions and ensures the correction only runs when Wave itself is the Snap.

### 2. Do not explicitly set XDG vars to empty strings
After fetching PAM env vars, any XDG key not present in PAM was left as `""` in `varsToReplace`, and was then passed to `shellutil.UpdateCmdEnv`. Depending on whether the key already existed in the command env, this either **unset** the variable or **set it to the empty string**, both of which break applications that rely on XDG Base Directory spec defaults (e.g. `XDG_RUNTIME_DIR` being empty caused the reported zsh-autosuggestions panic in uv/pipx-installed tools).

Added a filter that drops keys still holding empty values before calling `UpdateCmdEnv`. When a key is absent from the map, the child inherits the parent value (or uses spec defaults / shell rc files), which is the correct behavior on non-Snap systems and on Snaps where PAM didn't export that particular variable.

## Verification
- `go vet ./pkg/shellexec/` passes.
- Manually verified that with `SNAP` set but `SNAP_NAME != waveterm`, the Snap-correction block is no longer entered.
- Manually verified that with `SNAP_NAME=waveterm` and a partial PAM env, missing keys are omitted from `varsToReplace` rather than set to `""`.

Fixes #3336

### DIFF
--- a/pkg/shellexec/shellexec.go
+++ b/pkg/shellexec/shellexec.go
@@ -648,8 +648,12 @@ func StartLocalShellProc(logCtx context.Context, termSize waveobj.TermSize, cmdS
 	  overrides them to point to snap directories. We will get the correct values, if
 	  set, from the PAM environment. If the XDG variables are set in profile or in an
 	  RC file, it will be overridden when the shell initializes.
+
+	  Only apply this correction when Wave itself is running as a Snap (SNAP_NAME == "waveterm"),
+	  not when $SNAP is merely inherited from a parent Snap process (e.g. a CLI tool
+	  installed via Snap).
 	*/
-	if os.Getenv("SNAP") != "" {
+	if os.Getenv("SNAP_NAME") == "waveterm" {
 		log.Printf("Detected Snap installation, correcting XDG environment variables")
 		varsToReplace := map[string]string{"XDG_CONFIG_HOME": "", "XDG_DATA_HOME": "", "XDG_CACHE_HOME": "", "XDG_RUNTIME_DIR": "", "XDG_CONFIG_DIRS": "", "XDG_DATA_DIRS": ""}
 		pamEnvs := tryGetPamEnvVars()
@@ -659,6 +663,15 @@ func StartLocalShellProc(logCtx context.Context, termSize waveobj.TermSize, cmdS
 				if _, ok := varsToReplace[k]; ok {
 					varsToReplace[k] = pamEnvs[k]
 				}
+			}
+		}
+		// Remove keys that still have empty values so they are not explicitly set to
+		// empty strings in the child environment (which breaks tools that fall back to
+		// XDG spec defaults).  When a key is absent from varsToReplace the child will
+		// either inherit the parent value or rely on shell rc files / spec defaults.
+		for k, v := range varsToReplace {
+			if v == "" {
+				delete(varsToReplace, k)
 			}
 		}
 		log.Printf("Setting XDG environment variables to: %v", varsToReplace)
--- a/pkg/shellexec/shellexec.go
+++ b/pkg/shellexec/shellexec.go
@@ -580,6 +580,10 @@ func StartRemoteShellJob(ctx context.Context, logCtx context.Context, termSize w
 	return jobId, nil
 }
 
+// StartLocalShellProc starts a local shell process with the given command string,
+// terminal size, command options, and connection name. It configures the shell
+// environment (including correcting XDG directory variables when running as a Snap),
+// sets up a PTY, and returns a ShellProc instance managing the running process.
 func StartLocalShellProc(logCtx context.Context, termSize waveobj.TermSize, cmdStr string, cmdOpts CommandOptsType, connName string) (*ShellProc, error) {
 	if cmdOpts.SwapToken == nil {
 		return nil, fmt.Errorf("SwapToken is required in CommandOptsType")


Jason-Shen2:
@CLAassistant check

Jason-Shen2:
@CLAassistant check

Jason-Shen2:
Added docstrings for the new code in \`pkg/shellexec/shellexec.go\` in commit 79d1959a to address the docstring coverage warning.

Jason-Shen2:
@CLAassistant check

PR #3407, opened 2026-07-02
Preserve all windows/workspaces on close, not just the last one

Jordy098:
## Summary
- Terminal blocks already persist their cwd (`cmd:cwd` meta, set via the OSC 7 shell-integration handler) and restore it when their shell relaunches — but that only helps if the `Block`/`Workspace` survives in the DB.
- Closing a window while other windows are still open (or with `window:savelastwindow` off) always deleted that window's workspace/tabs/blocks on close, discarding any persisted cwd along with it. Only a single-window full app quit reliably survived a restart.
- Adds a new setting `window:restoreallwindows` (default `true`) so every window is preserved on close, not just the last one, letting all previously-open terminal paths come back after a full app restart.

## Test plan
- [ ] Open 2+ windows, `cd` to different directories in each terminal block and let the prompt render (fires OSC 7 → `cmd:cwd`)
- [ ] Close one window via the close button (not full quit) — its workspace should remain in the DB instead of being deleted
- [ ] Quit the app fully and relaunch
- [ ] Confirm all windows reopen with their terminals in the same working directory as before
- [ ] Repeat with `window:restoreallwindows` set to `false` to confirm the legacy single-window behavior still works


andrewinalaska:
Traced this through fully (not a bot finding — I've been working in this exact area for #3512-3514 and wanted to check for interaction effects). I think there's a real regression here, worth a look before merge.

**The issue:** `deleteAllowed` (set in the `close` handler, `emain-window.ts`) is the *only* thing that gates whether `WindowService.CloseWindow(windowId, true)` gets called from the `closed` handler. And `wcore.CloseWindow` (`pkg/wcore/window.go`) is the *only* code path in the entire repo that removes an entry from `client.WindowIds` — confirmed by grepping every reference to `client.WindowIds`; every other one only appends (on window creation) or reorders (on focus), never removes.

With this PR's new `preserveThisWindow` condition and `window:restoreallwindows` defaulting to `true`, `deleteAllowed` is essentially never set to `true` anymore for an ordinary window close (only a full app quit still hits the paths that eventually clean things up). That means `wcore.CloseWindow` never runs for a plain "click the X on one window while others stay open" action, which means:

1. That window's ID is never removed from `client.WindowIds`.
2. Its window/workspace DB rows are never deleted (`DeleteWorkspace`/`DBDelete` never run).
3. On every subsequent app launch, `relaunchBrowserWindows()` iterates `clientData.windowids` and faithfully reopens it — forever, since nothing else ever prunes that list.

So with the new default, manually closing a window doesn't really "close" it in the sense a user would expect — it just hides it until the next launch, and it keeps coming back indefinitely, even windows closed weeks ago. I don't think that's the intended behavior described in the PR body (which frames this around *app restart* restoring what was open, not "closed windows resurrect forever"). It might be worth distinguishing "window closed individually by the user" from "window still open when the app quit" — e.g. only preserving windows that are still open at actual app-quit time, rather than preserving every window that has ever been closed while `restoreallwindows` is on.

Happy to be wrong here if I'm missing some other cleanup path — but I traced every `client.WindowIds` mutation site and this was the only removal path I could find.

PR #3426, opened 2026-07-11
feat(wsh): add 'tab list' and 'tab move' commands for scriptable tab management

### LINKED ISSUE #3425: [Feature]: wsh command to reorder tabs (expose updateworkspacetabids)
### Use case
I colour-code my tabs by session activity (working / needs input / done) using `wsh setbg` + `wsh setmeta tab:flagcolor` driven by an external watcher. I'd like the watcher to also *sort* tabs so the ones needing attention are first — like browsers move audible tabs, or how pinned tabs group.

### What exists today
Drag-reorder in the UI is backed by the `updateworkspacetabids` RPC (`args: [workspaceId, [tabId, ...]]`), and `workspacelist` returns each workspace's current ordered `tabids`. Both work when called directly over `wave.sock` with the JWT handshake (tested on v0.14.5), but there's no CLI wrapper, so scripts must speak the socket protocol themselves.

### Proposal
```
wsh tab list                          # tab ids + names in current workspace, in order
wsh tab move <tabid> --index 0        # or: wsh tab order <id1> <id2> ...
```
A guard that rejects lists which drop/add tab ids would keep it safe.

Related: #3196 (wsh tab rename), #3285 (wsh tab create) — together these would make tabs fully scriptable.

Jason-Shen2:
## Summary

This PR adds two new wsh subcommands under `wsh tab` to make tabs scriptable from the command line, addressing #3425.

### New commands

**`wsh tab list`** (alias: `ls`) — lists tabs in the current workspace with their ids, names, and indices.
- Defaults to the current workspace from `WAVETERM_WORKSPACEID`, overridable with `--workspace <id>`
- Supports `--json` flag for machine-readable output
- Example:
  ```
  $ wsh tab list
  INDEX  TAB ID                                NAME
  0      550e8400-e29b-41d4-a716-446655440000  Terminal
  1      650e8400-e29b-41d4-a716-446655440001  google.com
  ```

**`wsh tab move <tabid> --index <n>`** — moves a tab to the specified 0-based position within its workspace.
- Validates that the tab exists in the workspace and that the index is in range
- Guards against dropping/adding tabs (the set of tab ids is preserved exactly)
- No-op when the tab is already at the target index
- Example:
  ```
  $ wsh tab move 650e8400-e29b-41d4-a716-446655440001 --index 0
  moved tab 650e8400-... from index 1 to 0
  ```

### Bug fix
- Fixed a small bug in `wsh workspace list` where the `workspaceId` field was being printed to stderr instead of stdout due to a `WriteStderr`/`WriteStdout` typo.

### Testing
- Added unit tests for the tab reordering logic (`reorderTabIds`) covering 8 scenarios: first↔last, left/right moves, adjacent swaps, and two-element arrays
- All existing tests in `cmd/wsh/cmd/` continue to pass
- `go build ./cmd/wsh/...` and `go vet ./cmd/wsh/...` clean
- `gofmt` verified

Closes #3425

### DIFF
--- a/cmd/wsh/cmd/wshcmd-tab.go
+++ b/cmd/wsh/cmd/wshcmd-tab.go
@@ -1,0 +1,226 @@
+// Copyright 2026, Command Line Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"text/tabwriter"
+
+	"github.com/spf13/cobra"
+	"github.com/wavetermdev/waveterm/pkg/waveobj"
+	"github.com/wavetermdev/waveterm/pkg/wshrpc"
+	"github.com/wavetermdev/waveterm/pkg/wshrpc/wshclient"
+)
+
+var tabCommand = &cobra.Command{
+	Use:   "tab",
+	Short: "Manage tabs",
+	Long:  "Commands for listing and reordering tabs.",
+}
+
+var tabListCmd = &cobra.Command{
+	Use:     "list",
+	Aliases: []string{"ls"},
+	Short:   "List tabs in a workspace",
+	Long: `List tabs with their ids and names, in their current order.
+Defaults to the current workspace (from WAVETERM_WORKSPACEID).`,
+	RunE:    tabListRun,
+	PreRunE: preRunSetupRpcClient,
+}
+
+var tabMoveCmd = &cobra.Command{
+	Use:   "move <tabid> --index <n>",
+	Short: "Move a tab to a new position in the tab bar",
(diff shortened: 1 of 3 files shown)


Jason-Shen2:
Addressed both CodeRabbit review comments in commit b9e0e680:

1. **Nil slice → null JSON**: Changed `var entries []tabListEntry` to `var entries []tabListEntry = make([]tabListEntry, 0)` so empty workspaces serialize as `[]` instead of `null`.
2. **MarkFlagRequired for --index**: Replaced the manual `cmd.Flags().Changed("index")` check with `tabMoveCmd.MarkFlagRequired("index")` in `init()` for standard Cobra enforcement.

PR #3478, opened 2026-08-18
fix: serialize concurrent mermaid renders in markdown preview

anik8das:
Fixes #3191.

Background is in that thread. My analysis and a position bisect are in [this comment](https://github.com/wavetermdev/waveterm/issues/3191#issuecomment-5334392713): a file with six blocks where block 6 was byte identical to block 1, in which blocks 1 and 2 rendered every time, block 6 never did, and the set changed between reloads. Identical content at a different position giving a different result is what ruled out syntax and pointed at a race.

## Problem

Each `<Mermaid>` component calls `mermaid.run()` from its own `useEffect`, so a markdown file with several diagrams starts every render in parallel. Two things go wrong.

**1. Colliding SVG ids.** `mermaid.run()` mutates module-global state and derives its element id from `Date.now()`:

```js
this.next = deterministic ? () => this.count++ : () => Date.now();
```

Calls that overlap, or land in the same millisecond, get the same `mermaid-<id>`. Mermaid scopes each diagram's `<style>` block by that id, so colliding ids let one diagram's styles apply to another.

**2. Repeated `initialize()`.** `initializeMermaid()` sets `mermaidInitialized = true` *after* `await import("mermaid")`. Every component that mounts before that import resolves gets past the guard, so they all call `mermaid.initialize()`, resetting global config while other renders are in flight.

### A. See the symptom

Save this as `repro.md` and preview it in Wave. On `main` the diagrams overlap or fail to appear, and which ones break changes between reloads. With this PR all four render in place.

````markdown
# repro

```mermaid
flowchart TD
  A[Start] --> B{Decision}
  B -->|Yes| C[Action 1]
  B -->|No| D[Action 2]
```

```mermaid
sequenceDiagram
  participant Alice
  participant Bob
  Alice->>Bob: Hello
  Bob-->>Alice: Hi back
```

```mermaid
flowchart LR
  X[Input] --> Y[Process] --> Z[Output]
```

```mermaid
flowchart TD
  P[One] --> Q[Two]
  Q --> R[Three]
```
````

### B. Measure the id collision

This needs no new dependencies. Save it at the repo root as `mermaid-race-repro.html` (after `npm install`) and open it in a browser. It renders 8 diagrams the way `markdown.tsx` does, then the way this PR does, and prints how many ended up sharing an SVG id.

<details>
<summary>mermaid-race-repro.html</summary>

```html
<!doctype html><meta charset="utf-8">
<title>mermaid concurrent render repro</title>
<style>body{font:14px system-ui;margin:24px}pre{background:#f4f4f5;padding:12px;border-radius:6px}#diagrams{display:none}</style>
<h3>mermaid concurrent render repro</h3>
<pre id="out">running...</pre>
<div id="diagrams"></div>
<script src="./node_modules/mermaid/dist/mermaid.min.js"></script>
<script>
const N = 8, TRIALS = 6;
const log = [];
const make = i => {
  const el = document.createElement("div");
  el.textContent = `flowchart TD\n  A${i}["a${i}"] --> B${i}["b${i}"]`;
  document.getElementById("diagrams").appendChild(el);
  return el;
};
// the queue this PR adds
let queue = Promise.resolve();
const runQueued = (node, text) => {
  const r = queue.then(() => { node.removeAttribute("data-processed"); node.textContent = text; return mermaid.run({ nodes: [node] }); });
  queue = r.catch(() => {});
  return r;
};
(async () => {
  mermaid.initialize({ startOnLoad: false, theme: "dark", securityLevel: "strict" });
  for (const queued of [false, true]) {
    let worstDupes = 0, worstRendered = N;
    for (let t = 0; t < TRIALS; t++) {
      document.getElementById("diagrams").innerHTML = "";
      const els = Array.from({ length: N }, (_, i) => make(i));
      const texts = els.map(e => e.textContent);
      // every <Mermaid> useEffect fires at once, as in markdown.tsx
      await Promise.all(els.map((e, i) => queued ? runQueued(e, texts[i]) : mermaid.run({ nodes: [e] })));
      const ids = els.map(e => e.querySelector("svg")?.id).filter(Boolean);
      worstDupes = Math.max(worstDupes, ids.length - new Set(ids).size);
      worstRendered = Math.min(worstRendered, els.filter(e => e.querySelector("svg")).length);
    }
    log.push(`${queued ? "queued (this PR)  " : "concurrent (before)"}  worst duplicate ids: ${worstDupes}/${N}   rendered: ${worstRendered}/${N}`);
    document.getElementById("out").textContent = log.join("\n");
  }
})();
</script>
```

</details>

Output here, on this repo's mermaid 11.15.0:

```
concurrent (before)  worst duplicate ids: 6/8   rendered: 8/8
queued (this PR)     worst duplicate ids: 0/8   rendered: 8/8
```

### What these do and do not show

B measures the id collision deterministically and runs in any browser. A shows the user-visible symptom, but I have only been able to observe it in Wave. In headless Chromium all eight diagrams still render even with fully concurrent calls, so the overlap and the missing diagrams appear to need the app's markdown preview specifically. That is why B measures the underlying id collision rather than the visual result: it is the part of the failure I can demonstrate anywhere.

## Fix

Queue the `run()` calls so only one executes at a time, and memoize the init promise instead of setting a flag after the await. No change in behavior for single-diagram documents.

The node's text is written inside the queued task rather than before queueing, because the node contents are what `mermaid.run()` reads when the task executes. Writing them earlier would leave a window where a later effect on the same node could replace the text an earlier queued render was about to read.

## Note on `deterministicIds`

The issue suggests `deterministicIds: true` as a possible fix. It makes things worse. The id generator is constructed inside each `run()` call, so with one node per call every diagram gets `mermaid-0`:

| deterministicIds | ids from 3 sequential `run()` calls |
| --- | --- |
| `false` | `mermaid-1787091579824`, `mermaid-1787091579860`, `mermaid-1787091579867` |
| `true` | `mermaid-0`, `mermaid-0`, `mermaid-0` |

## Testing

- `prettier --check` passes on the changed code. The one remaining warning is a pre-existing import ordering difference that is also present on `main`, so I left it alone rather than mix an unrelated reformat into this PR.
- `tsc --noEmit` reports the same 17 pre-existing errors as the base commit, none in `markdown.tsx`.

## Deliberately out of scope

`useEffect` has no cleanup, so a queued render for an unmounted component still runs and can call `setError`/`setIsLoading` afterwards. That is pre-existing and separate from the concurrency bug, so I left it out to keep this to one logical change. Happy to follow up in another PR if you want it.


anik8das:
Addressed the review feedback in c3d805c: the node's text is now written inside the queued task instead of before it, so a render always reads its own chart. That was the substantive half of the finding, and it closes a window the queue had widened.

I left the `useEffect` cleanup out on purpose. A queued render belonging to an unmounted component can still call `setError`/`setIsLoading` afterwards, but that is pre-existing and separate from the concurrency bug, and CONTRIBUTING.md asks for one logical change per PR. Happy to do it as a follow-up if you want it.

I also added runnable reproduction steps to the description: a markdown file that shows the symptom in the preview, and a standalone HTML file that measures the SVG id collisions directly with no extra dependencies.

The earlier merge-gatekeeper failure was a timing artifact. It timed out waiting on `license/cla` before the CLA was signed, and it passes on the new commit.

PR #3486, opened 2026-08-31
Fix OSC 8 links opening in external browser

silexKhan:
Fixes #3165

## Problem
The xterm.js default OSC 8 handler confirms the URL, then calls `window.open()` without the URL. Wave’s Electron window-open policy denies that blank window, so clicking OK does not open a browser.

## Fix
Provide xterm.js with a link handler that delegates to Wave’s existing `openLink(uri)` path. This preserves `web:openlinksinternally`; when it is false, the URL is opened with the OS default browser.

## Verification
- `npm run build:prod`
- `npx vitest run frontend/app/view/term/osc-handlers.test.ts`
- `git diff --check`

`npx tsc --noEmit` still reports existing unrelated preview mock/type errors; none are in the changed file.


Tengs-Fan:
could anyone merge this ? I had the smae problem

PR #3495, opened 2026-09-04
Claude/wave terminal localization rcp9nn

nodedropweb:


nodedropweb:
it was never planed to publish it here but i clicked the wromg button. That is not ready so i cloes this.

PR #3498, opened 2026-09-09
Fix PowerShell encoding for Windows environment variable names

### LINKED ISSUE #3481: pwsh: encodeEnvVarsForPowerShell rejects Windows-standard env var names like ProgramFiles(x86), breaking wsh token on every pwsh block
### Description

`encodeEnvVarsForPowerShell` in `pkg/util/shellutil/tokenswap.go` validates every environment variable name against the strict POSIX-identifier regex `^[A-Za-z_][A-Za-z0-9_]*$` (`IsValidEnvVarName` in `pkg/util/shellutil/shellquote.go`), and this validation is applied unconditionally to ALL shells in `EncodeEnvVarsForShell`, including `pwsh`.

Windows always defines environment variables containing parentheses, e.g. `ProgramFiles(x86)` and `CommonProgramFiles(x86)`. These names fail the regex, so `encodeEnvVarsForPowerShell` returns:

```
invalid env var name: "ProgramFiles(x86)"
```

`EncodeEnvVarsForShell` propagates this as a hard error, which aborts the entire token exchange, not just the one offending variable.

### Why this matters

This is not a cosmetic edge case, it fires on essentially every PowerShell block launch on Windows. Wave's own bundled `pwsh` shell-integration script (`wavepwsh.ps1`, generated for the `pwsh` shell type) runs:

```powershell
wsh token $env:WAVETERM_SWAPTOKEN pwsh 2>$null
```

The `2>$null` silently swallows the error, so the failure is invisible in normal use, but the entire env-var/init-script exchange for that block never runs.

### Root cause

`$env:ProgramFiles(x86) = "..."` is invalid PowerShell syntax (parens are parsed as a command invocation), so the strict validator was presumably written to avoid emitting that. But PowerShell has a second, valid assignment form for exactly this case:

```powershell
${env:ProgramFiles(x86)} = "..."
```

The `${env:NAME}` delimited form tolerates parentheses (and most other characters) in the variable name. The validator doesn't need to reject these names for PowerShell at all, it just needs to emit the delimited form instead of the bare `$env:NAME` form.

### Suggested fix

Scope the fix to `encodeEnvVarsForPowerShell` only (leave `encodeEnvVarsForBash` / `encodeEnvVarsForFish` untouched, the POSIX identifier rule is correct there):

- Emit `${env:%s} = %s` instead of `$env:%s = %s`.
- Only reject a name if it's structurally impossible to express in that form (empty, or contains a literal `}`), rather than requiring a POSIX identifier.

### Environment

- Wave Terminal / wsh version: v0.14.5 (also present in current `main` as of this report, `pkg/util/shellutil/tokenswap.go` is unchanged on that path)
- OS: Windows 11
- Shell: pwsh (PowerShell 7)

### Repro

```go
package main

import (
	"fmt"
	"github.com/wavetermdev/waveterm/pkg/util/shellutil"
)

func main() {
	env := map[string]string{"ProgramFiles(x86)": `C:\Program Files (x86)`}
	_, err := shellutil.EncodeEnvVarsForShell("pwsh", env)
	fmt.Println(err) // invalid env var name: "ProgramFiles(x86)"
}
```

Or from a shell with any real Windows environment:

```
wsh token <valid-swap-token> pwsh
# Error: error encoding env vars: invalid env var name: "ProgramFiles(x86)"
```

vortsghost2025:
## Summary

Fixes #3481.

- Emit braced `${env:NAME}` assignments for PowerShell environment variables.
- Accept Windows-standard names such as `ProgramFiles(x86)` and `CommonProgramFiles(x86)`.
- Reject empty names and names containing `}`.
- Preserve Bash/fish validation and add coverage for ordinary `PATH` plus Windows names.

## Test plan

- `go test ./pkg/util/shellutil -count=1`
- `go test ./pkg/util/shellutil -run 'TestEncodeEnvVarsForPowerShell' -count=1`
- PowerShell 7.6.5 runtime validation


vortsghost2025:
Addressed the open PowerShell review finding in follow-up commit `d9fee7c26babe0c7fc593abc186e7888ef77f7c8`. PowerShell environment-variable names containing backticks are now rejected alongside closing braces, and the regression test covers `FOO\``. The PR head is updated; GitHub’s CLA and CodeRabbit checks are currently pending.

PR #3512, opened 2026-09-16
fix(shellexec): preserve argv boundaries for SSH and WSL command launches

andrewinalaska:
## Summary
`StartRemoteShellProc`/`StartRemoteShellProcNoWsh` and the WSL launch path built the remote command by flattening `shellOpts` (e.g. `["-c", cmdStr]`) into a single string via `strings.Join`, then sending that string down the raw SSH exec channel (or `sh -c` for WSL). The remote login shell re-parses that string itself, and `-c` only binds its *next single token* as the script body — so any multi-word `cmdStr` silently loses its trailing arguments as unused positional parameters.

Concretely, `wsh run -- tmux attach -t mysession` over SSH would flatten to something like `bash -c tmux attach -t mysession`. The remote shell binds `-c` to `tmux` alone; `attach`, `-t`, `mysession` become `$0`/`$1`/`$2` and are dropped. The result is a bare `tmux` invocation instead of attaching to the named session.

Local exec was never affected — it already used structured `exec.Command(shellPath, shellOpts...)`.

## Fix
- New `SerializeCommandForShell` (`pkg/util/shellutil/shellquote.go`) is now the single place argv is ever flattened into shell text, one code path per outer-shell dialect (POSIX/fish/pwsh).
- Outer-shell type (the remote login shell for SSH; hardcoded POSIX for WSL, since it invokes `sh -c`) is now distinguished from inner shell type.
- Removed scattered compensating-quote hacks: the `shellPath = "& " + shellPath` PowerShell call-operator hack, and a hand-embedded literal-quote hack in fish's `-C` arg (now uses `HardQuoteFish`).
- WSL paths switched from tilde-relative to absolute (`remoteInfo.HomeDir`-based) paths.
- Centralized, deterministic (sorted) environment-variable encoding instead of unquoted ad hoc string prepends.
- Fixed `StartRemoteShellProcNoWsh` / the WSL no-wsh fallback, which previously ignored a non-empty `cmdStr` entirely and always launched an interactive shell.
- Fixed a secrets-in-logging issue: the swap token and JWT were logged raw (`conn.Infof("swaptoken: %s", ...)`, and a WSL `log.Printf` of the fully assembled command including env). Replaced with a `describeForLog`-style helper that logs only shell types, executable, argc, and env key *names* — verified by a dedicated test asserting token/JWT values never appear in log output.

## Testing
- Unit tests for the new serializer across shell dialects.
- A 30s fuzz run of `SerializeCommandForShell` round-tripped through a real shell (83,230 execs, 0 failures).
- An in-process integration test using a real `golang.org/x/crypto/ssh` server/client (same wire protocol and same `ssh.Session.Start` call production code uses) that reproduces the bug pre-fix and confirms the fix post-fix.
- A real-tmux end-to-end test over the same SSH harness, reproducing the exact `wsh run -- tmux attach` failure shape: old construction shows a `command not found`-style corrupted invocation, new construction attaches cleanly.

Zero changes under `cmd/`.


andrewinalaska:
Addressed the codex and coderabbit findings in e0171169:
- Fixed the `HardQuotePowerShell` newline-doubling bug (P2), added a regression test.
- Removed/redacted the remaining raw `cmdStr`/`shellOpts`/packed-swaptoken debug logs that predated `describeForLog` and were missed by it (`StartWslShellProc`, `StartRemoteShellProc`, `StartRemoteShellJob`, `StartLocalShellProc`) — `StartRemoteShellJob` now routes through `describeForLog` too.
- `TestNoWsh_SessionWrapDeliversCmdStrOverRealSSH` now actually calls `SessionWrap.Start()` and captures output via a buffer, instead of calling `session.Output` directly.
- Fixed the two QF1012 staticcheck findings in `tokenswap.go`.
- Replaced the tmux e2e test's fixed sleeps with a deterministic settle-sentinel + poll (stable over 8 consecutive local runs, ~5x faster).

andrewinalaska:
@codex review

PR #3513, opened 2026-09-16
fix(blockframe-header): show frame:title on durable shell blocks

andrewinalaska:
## Summary
`frame:title` block meta is read and rendered by every block header type except durable shell-controller blocks. `BlockFrame_Header` gates its entire viewName/title render branch on `!useTermHeader`, and durable shell blocks (`controller: shell` with `jobcontroller.IsBlockIdTermDurable` true) always have `useTermHeader === true` — so that whole branch, and `frame:title` along with it, is skipped unconditionally for this block type.

Practical effect: every durable terminal pane on a given connection shows an identical, generic connection-name header, with no way to visually distinguish them — for example several tmux-attached sessions opened on the same remote host via `wsh createblock`/`wsh setmeta frame:title=...` all look the same in the tab/pane header, even though setting `frame:title` succeeds and is visible via `wsh getmeta`.

## Fix
Render `frame:title` inside the `useTermHeader` branch too, next to the connection button, so a caller can label a durable session distinctly (e.g. by the tmux session name it's attached to).

## Testing
Manually verified against a locally-built app: created a durable shell block via `wsh createblock term controller=shell connection=<conn> cmd=... frame:title=<label>`, confirmed the header now shows `<label>` instead of only the connection name.


andrewinalaska:
Addressed the codex/coderabbit overflow finding in 7f5500a1: the new frame:title element wasn't actually inside the DOM scope the reused \`block-frame-view-type\` class's shrink/truncate rules apply to (that class is a descendant selector under \`block-frame-default-header-iconview\`), so a long title could overflow and push the durable badge/end controls out of view. Added a dedicated \`block-frame-durable-title\` rule scoped to this element's actual position.

andrewinalaska:
@codex review

PR #3514, opened 2026-09-16
fix(wcore): add ListAllWorkspaces so wsh CLI sees unsaved/scratch workspaces

andrewinalaska:
## Summary
`wsh workspace list` and `wsh blocks list`'s default "all workspaces" enumeration are built on `wcore.ListWorkspaces`, which excludes any workspace missing `Name`, `Icon`, or `Color`. That's intentional for its other callers: the frontend workspace switcher (`WorkspaceService.ListWorkspaces`) and, as caught in review, the Electron Workspace menu (`emain-menu.ts`) and `Alt+Ctrl+<N>` workspace-switch shortcuts (`emain-tabview.ts`) — all three rely on unsaved (scratch) workspaces staying excluded, since `CreateWindow` deliberately creates one of those for a new window (`CreateWorkspace(..., applyDefaults=false)`) and `DeleteWorkspace` auto-cleans it up on close unless it's since been named.

The bug is that the exact same exclusion also blinds the **CLI** to those workspaces — and everything inside them. Confirmed against a real user's database: a workspace holding their own long-running Claude Code session's active tab had never been named, and was completely invisible to `wsh blocks list` the entire time it was in continuous daily use.

## Fix (revised twice through review)
- `wcore.ListWorkspaces` is completely unchanged — still excludes unsaved workspaces, still shared by the frontend switcher.
- New `wcore.ListAllWorkspaces` includes them too, without ever writing anything to the database.
- **A prior revision of this PR pointed the shared `WorkspaceListCommand` RPC at the new function**, on the claim that it was CLI-only — codex correctly caught that this RPC is also called from `emain-menu.ts` and `emain-tabview.ts` (Electron main-process code, a call-site my original search missed entirely), which would have added blank scratch-workspace entries to the Workspace menu and shifted keyboard-shortcut indices. Fixed by adding a **separate** `WorkspaceListAllCommand` RPC (regenerated through the real codegen: `cmd/generatego`, `cmd/generatets`) and pointing only `wsh workspace list` / `wsh blocks list` at it. `WorkspaceListCommand` itself, and every one of its other callers, is untouched.

## Test plan
`pkg/wcore/workspace_test.go` — real (non-mocked) sqlite-backed tests via `wstore.InitWStore()`:
- `ListWorkspaces` excludes an unsaved workspace and never mutates it.
- `ListAllWorkspaces` includes it too, also without mutating anything.

```
go test ./pkg/wcore/... ./pkg/wshrpc/...
```
Also verified a full `npm run build:prod` (electron-vite, compiles main/preload/renderer together) to confirm `emain-menu.ts`/`emain-tabview.ts` build clean and unaffected against the regenerated bindings.


andrewinalaska:
All three findings here were correct and pointed at a real design conflict, not just implementation nits — addressed in 432be083 with a redesign rather than a patch:

- **codex P1 (unsaved workspace lifecycle)**: confirmed — `CreateWindow` deliberately creates unnamed scratch workspaces (`applyDefaults=false`), and `DeleteWorkspace`/the frontend switcher both depend on blank Name/Icon meaning "unsaved". The fix no longer touches or persists anything into any workspace. `ListWorkspaces` keeps its exact original behavior (switcher-facing, excludes unsaved); a new `ListAllWorkspaces` includes them too, purely by not filtering — never by writing.
- **codex P1 (atomicity)**: moot now — there's no write at all, so no read-then-write race is possible.
- **coderabbit (swallowed error)**: moot for the same reason.

New tests in `pkg/wcore/workspace_test.go` (real sqlite via `wstore.InitWStore()`, not mocked) assert both functions leave an unsaved workspace's stored fields completely untouched, in addition to checking inclusion/exclusion.

andrewinalaska:
@codex review

andrewinalaska:
Good catch — my "never called from the frontend" claim in the last revision was wrong. My call-site search covered pkg/, frontend/, and cmd/, but missed emain/ (Electron main-process TS) entirely, a third source tree in this repo. Confirmed both call sites directly (\`emain-menu.ts:49\`, \`emain-tabview.ts:76\`) and their reliance on unsaved workspaces staying excluded (menu labels, positional shortcut indexing).

Fixed in 9fb4a55e by NOT touching the shared \`WorkspaceListCommand\` RPC at all — it's back to its exact original behavior. Added a separate \`WorkspaceListAllCommand\` RPC instead (generated via the real codegen, not hand-written), and pointed only \`wsh workspace list\`/\`wsh blocks list\` at it. Verified with a full \`npm run build:prod\` (electron-vite, compiles main/preload/renderer together) that both emain files still build clean and are completely unaffected.

andrewinalaska:
@codex review

